PR #29, opened 2016-08-10
Use new admin page module

LukasReschke:
Makes this page a single page as implemented in https://github.com/nextcloud/server/pull/796


blizzz:
still works \o/

PR #34, opened 2016-09-10
Replace generic icon with lock icon from password_policy

eppfel:
I think it is a more approriate symbol.
@nextcloud/designers


MariusBluem:
Saml is a user backend ... For that reason, i like the current Icon :grin:

MorrisJobke:
Yes. This is a user backend and should stay the people icon.

Issue #23, opened 2016-07-08
MS ADFS guide


MichaelHeimann:
Hi,

I've seen you support SAML and immediately wanted to enable it - but failed.

I've tried to configure MS ADFS 3 as IDP with user_saml as SP and got the redirect working and as far as I can see the authentication was granted from the server but I only see "invalid_response Not authenticated" in the browser.

Help like how to configure the claim rule on an ADFS setup would be appreciated.
e.g. http://www.cisco.com/c/en/us/support/docs/unified-communications/unified-communications-manager-callmanager/118771-configure-samlsso-00.html

Also enabling to import IDP metadata to set config values correctly would be nice.

Cheers,
Michael

LukasReschke:
Hi Michael

Thanks for giving the app a try. While I didn't test ADFS I'm pretty sure it should work, can you screenshot me your settings? I can then try to test it locally as well and document that properly :-)

Thanks a lot,
Lukas

MichaelHeimann:
Hi Lukas

thanks for your response. Which settings would you be interested in?

This is how the user_saml settings in nextcloud look like.
×

​
I am getting redirected to the ADFS login page, but after authenticating I
only get :
"

invalid_response

Not authenticated

"

:(

This is what I can do on ADFS to get attributes in the SAML info from ADFS:


×

​

This is what SAML tracer addon in firefox gives me:


<samlp:Response ID="_dbde8760-bbc1-4c6f-9bc2-4580f41f4e9c"
                Version="2.0"
                IssueInstant="2016-07-10T20:03:37.657Z"

Destination="https://share.avodaq.com/index.php/apps/user_saml/saml/acs"
                Consent="urn:oasis:names:tc:SAML:2.0:consent:unspecified"
                InResponseTo="ONELOGIN_79ce41b9d4eaef90d6f09d6ac3c6eeec1f221577"
                xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
                >
    <Issuer xmlns="urn:oasis:names:tc:SAML:2.0:assertion">http://adfs.avodaq.com/adfs/services/trust</Issuer>
    <samlp:Status>
        <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success" />
    </samlp:Status>
    <Assertion ID="_5ef66942-f4ee-48f5-9c3c-0a4ffcf020af"
               IssueInstant="2016-07-10T20:03:37.656Z"
               Version="2.0"
[...]

PR #38, opened 2016-09-30
Explicitly set authncontext

LukasReschke:
Should help with ADFS

Ref https://github.com/nextcloud/user_saml/issues/23#issuecomment-247874576

### DIFF
--- a/lib/samlsettings.php
+++ b/lib/samlsettings.php
@@ -57,6 +57,10 @@ class SAMLSettings {
 				'wantNameId' => ($this->config->getAppValue('user_saml', 'security-wantNameId', '0') === '1') ? true : false,
 				'wantNameIdEncrypted' => ($this->config->getAppValue('user_saml', 'security-wantNameIdEncrypted', '0') === '1') ? true : false,
 				'wantXMLValidation' => ($this->config->getAppValue('user_saml', 'security-wantXMLValidation', '0') === '1') ? true : false,
+				'requestedAuthnContext' => [
+					\OneLogin_Saml2_Constants::AC_PASSWORD_PROTECTED,
+					\OneLogin_Saml2_Constants::AC_WINDOWS,
+				],
 			],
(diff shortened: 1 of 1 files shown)


MichaelHeimann:
@LukasReschke Thanks, that works great and also fixes a problem with external access which I discovered.

LukasReschke:
I'll merge that then once I've added some automatic integration tests for Shibboleth to ensure it won't add any regressions.

rzmf:
Why not setting authncontext = false?  if the idp allows login via some obscure mechanism it is up to the idp to allow login or not.

        'requestedAuthnContext' => false,

maybe also setting        'requestedAuthnContextComparison' => 'minimum'  would  be necessary

Ollienator:
I agree with rzmf.  'requestedAuthnContext' should be set to false instead of explicitely requesting a set of authentication contexts.  If any authentication contexts are requested, it will override the authentication method settings configured on the ADFS server (e.g. X509, Windows, Forms).

See this comment from the onelogin repo:
```
        // Authentication context.
        // Set to false or don't present this parameter and no AuthContext will be sent in the AuthNRequest,
        // Set true and you will get an AuthContext 'exact' 'urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport'
        // Set an array with the possible auth context values: array ('urn:oasis:names:tc:SAML:2.0:ac:classes:Password', 'urn:oasis:names:tc:SAML:2.0:ac:classes:X509'),
        'requestedAuthnContext' => true,
```

LukasReschke:
Fair enough. Will test with Shibboleth and OneLogin later as I'm anyways working on another feature that needs regression testing and if that works we can go the `requestedAuthnContext => false` route.

I won't have tested this against ADFS myself though. So I trust you, that you tested this 😉

LukasReschke:
Will be in 11.0

PR #61, opened 2016-12-23
add displayname

rzmf:


lib/Controller/SAMLController.php
```diff
@@ -85,6 +85,8 @@ public function __construct($appName,
 	 */
 	private function autoprovisionIfPossible(array $auth) {
 		$uidMapping = $this->config->getAppValue('user_saml', 'general-uid_mapping');
+		$displayName = $auth['urn:oid:2.5.4.42'][0] . ' ' . $auth['urn:oid:1.2.40.0.10.2.1.1.261.20'][0];
```
LukasReschke: From my understanding those should be configurable values – right?

rzmf: good point.
Two values like displayname.firstpart, displayname.secondpart to leave the admin a choice if first,last or last,first?

lib/Controller/SAMLController.php
```diff
@@ -85,13 +85,34 @@ public function __construct($appName,
 	 */
 	private function autoprovisionIfPossible(array $auth) {
 		$uidMapping = $this->config->getAppValue('user_saml', 'general-uid_mapping');
+		$displaynamefirstpartMapping = $this->config->getAppValue('user_saml','general-displaynamefirstpart_mapping');
+		$displaynamesecondpartMapping = $this->config->getAppValue('user_saml','general-displaynamesecondpart_mapping');
```
LukasReschke: I personally would prefer to not separate between first name and second name here. Rather just a single input field that can take multiple fields. (such as `{{fieldBar}} {{fieldFoo}}`) 

If you can give me commit access to your fork I can try to come up with something like that later.

rzmf: I gave you commit Access (i hope...)

LukasReschke: Yes. I just finished the automated integration test suite for Shibboleth. I'll see if I can find some time to dig into this here. :)

rzmf: btw: already got mailadress working, not ready for commit yet....

bne86:
Hi,
i was able to [implement displayname and email mapping](https://github.com/bne86/user_saml/commit/9b9887eb30d2540d69f40ed418922be7116e5878), they show up in a updated database schema.
But, beside of the fact that i extract a displayname and email from the SAML response, i can not see them in the WebUI and it looks like  [getDisplayName](https://github.com/bne86/user_saml/blob/stable11/lib/userbackend.php#L241) with some debug output is never called.

Any hints?

LukasReschke:
@bne86 From the look at your code I'd say this is because your backend doesn't define https://github.com/nextcloud/user_saml/pull/61/files#diff-eec818f31669ac6d6b4d1161cc38794bR113

bne86:
Thanks for the pointer, now i am getting the Log output and a error with the query that i can fix.
`OC_USER_BACKEND_GET_DISPLAYNAME` is deprecated, nevertheless it works and maybe i will find the corresponding not-deprecated functionality.

lib/Controller/SAMLController.php
bne86: Any update on this mailaddress item? I am storing the mail address(preferences database, as email configkey for given uid, like LDAP is doing this too), but it is not shown in the users personal page. Also, i have configured the displayname and its shown in the upper right side of the WebUI, but also not on the personal page. Somewhat strange.

rzmf: Displayname works for me everywhere: upper right side, personal page, search field for sharing items.

regarding mailaddress: i have not found time to implement this better.

My current solution is: writing the email the moment the user gets created, no update, no changes:

SAMLController.php
```
$mailMapping = "urn:oid:0.9.2342.19200300.100.1.3";

                $mail='';
                if(isset($auth[$mailMapping])) {
                        if(is_array($auth[$mailMapping])) {
                                $mail = $auth[$mailMapping][0];
                        } else {
                                $mail = $auth[$mailMapping];
                        }
                }`

```

```
if(!$userExists && !$autoProvisioningAllowed) {
                                throw new NoUserFoundException();
                        } elseif(!$userExists && $autoProvisioningAllowed) {
                                $this->userBackend->createUserIfNotExists($uid,$displayName);
                                $user = $this->userManager->get($uid);
                                error_log("set email");
                                if (!is_null($user)) {
                                        $user->setEMailAddress($mail);
                                }
                                error_log("set email");
                                return;
                        }
```

bne86: Thank you very much for your response!

LukasReschke:
As discussed out-of-band with @bne86 I'll look into getting this into shape and merged into master soonish.

LukasReschke:
Let's close this in favour of https://github.com/nextcloud/user_saml/pull/90 where I also added integration tests.

@rzmf @bne86 Can I ask you to give it a test? Any feedback appreciated. Note that you also require https://github.com/nextcloud/server/pull/3507

PR #94, opened 2017-03-10
fixing some issues with getAttributeValue from saml response

bne86:


blizzz:
i think this was  solved already differently since, closing thus.

PR #98, opened 2017-03-13
Alternative logins

bne86:
Only as starting point:
Provide way to allow several logins at the same time, for example the integrated Nextcloud user database and SAML. Or LDAP and SAML...
For this, utilize the "Alternative Logins" part of the login screen

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -43,7 +43,8 @@ $userBackend = new \OCA\User_SAML\UserBackend(
 	$urlGenerator,
 	\OC::$server->getSession(),
 	\OC::$server->getDatabaseConnection(),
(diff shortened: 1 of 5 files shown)


bne86:
Some updates here? Currently this is a show stopper for us.

bne86:
Some updates here? Currently this is a show stopper for us.

cellisten:
Any progress? In #126 @LukasReschke mentions that he might have time to do it in September. I have a use case where we need to allow external users to login with local accounts rather than SSO.

schiessle:
solved with this pr https://github.com/nextcloud/user_saml/pull/196 and released in v1.5.0

PR #111, opened 2017-05-22
Update version on master

nickvergessen:


rullzer:
CI fails

nickvergessen:
>   [Exception]                                                                                                        
App "SSO & SAML authentication" cannot be installed because it is not compatible with this version of the server. 

But I don't get it, why it only fails on integration tests and all others can install the app just fine.

nickvergessen:
cc @LukasReschke if you have time

LukasReschke:
Master will work on 12 and 13 via the appstore and there are also integration tests in place which test against 12 and 13. Thus closing :)

PR #129, opened 2017-07-13
Fix SingleLogout with Nextcloud 12

brenard:
Hello,

Since Nextcloud 12, Single Logout not working : the logout interface link not using any more the user backend _getLogoutAttribute()_ method. To handle logout, we have to use _logout_ hook to trigger the _SLO_ request. Moreever, the _SLO Response_ was not correctly handle by using _php-saml processSLO()_ method.

My fix implement a trigger on _logout_ hook to redirect user (if it's logged by using _php-saml_) to new logout URL manage by _SAMLController_. This URL make the _SLO Request_ and the old _singleLogoutService_ now handle the _SLO Response_ by calling _php-saml processSLO()_ method and redirect user to standard Nextcloud _logout_ URL.

I also fix another bug regarding SAML standard : In the _SLO Request_, the _NameID_ element must be exactly the same as provided by the SSO server on the _SSO Response_. I personally use [Authentic2 SSO Server](https://dev.entrouvert.org//projects/authentic) which is really strict about that. To fix it, I made some changes on _php-saml_ library to permit to set _nameId NameQualifier_ attribute in _SLO Request_. This changes have been proposed and merged upstream (https://github.com/onelogin/php-saml/pull/226). In _user_saml_, I made some changes to retrieve and store in session the _nameId NameQualifier_ attribute's value of the _SSO Response_ and pass it on _SLO Request_.

### DIFF
--- a/3rdparty/vendor/onelogin/php-saml/lib/Saml2/Auth.php
+++ b/3rdparty/vendor/onelogin/php-saml/lib/Saml2/Auth.php
@@ -34,6 +34,14 @@ class OneLogin_Saml2_Auth
      */
     private $_nameidFormat;
 
+
+    /**
+     * NameID NameQualifier
+     *
+     * @var string
+     */
+    private $_nameidNameQualifier;
+
     /**
      * If user is authenticated.
      *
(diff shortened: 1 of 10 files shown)


LukasReschke:
Single logout is now added back with https://github.com/nextcloud/user_saml/pull/145

PR #131, opened 2017-07-14
Support the ability to parse IdP XML metadata

apicht:
This enhances the SAML plugin with the ability to provide IdP metadata XML, reducing the amount of configuration required in certain environments.

Changes:
- Update php-saml to 2.10.7 (IdP metadata parsing was added in 2.10.6)
- Add config values to the admin interface to support either pasting XML or URL to IdP metadata.

Implementation details:
- If a user provides both a URL and XML, the XML will be parsed and injected, followed by the metadata from the URL. This way the metadata provided by the IdP takes precedence and things like certificates should be kept up to date automatically.
- If the user specifies the IdP Entity ID or SSO URL, these will take precedence over values received from metadata.
- If the user specifies the IdP Entity ID, it must match the IdP metadata else the metadata is ignored.

Opportunities for improvement:
- Cache metadata in case URL becomes unavailable (since metadata is typically delivered by same IdP server that would handle the SAML request, this is probably generally moot.)
- Populate admin interface with values parsed from metadata to more clearly indicate to Nextcloud admin what exactly is being set by the metadata.

Tested with Okta.

### DIFF
--- a/3rdparty/composer.lock
+++ b/3rdparty/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "onelogin/php-saml",
-            "version": "2.10.5",
+            "version": "2.10.7",
             "source": {
                 "type": "git",
                 "url": "https://github.com/onelogin/php-saml.git",
-                "reference": "3319d7707f342e38291eee6b01a4a5f8df1b333b"
+                "reference": "b41f731803f303879f50fa76276908088197731c"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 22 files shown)


schiessle:
@apicht we are updated now to php-saml 3.0... Would you be interesting in rebasing your PR and update it? Thanks a lot!

LukasReschke:
Closing due to inactivity. If there's any interest in reviving this, we'd be happy to take a look :)

PR #163, opened 2017-10-17
Use authoriative autoloader

rullzer:
So we don't try to load other classes with this autoloader.


MorrisJobke:
Looks good 👍

PR #174, opened 2018-01-16
add a meaningful error message in case a empty uid is given

schiessle:
In some wired setup and with some attributes it might happen that the IdP doesn't send a valid uid to the Nextcloud server

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -99,6 +99,12 @@ class SAMLController extends Controller {
 				$uid = $auth[$uidMapping];
 			}
 
+			// make sure that a valid UID is given
+			if (empty($uid)) {
+				$this->logger->error('Uid "' . $uid . '" is not a valid uid please check your attribute mapping', ['app' => $this->appName]);
+				throw new \InvalidArgumentException('No valid uid given, please check your attribute mapping. Given uid: ' . $uid);
+			}
+
 			$userExists = $this->userManager->userExists($uid);
 			$autoProvisioningAllowed = $this->userBackend->autoprovisionAllowed();
 			if($userExists === true) {


lib/Controller/SAMLController.php
```diff
@@ -99,6 +99,12 @@ private function autoprovisionIfPossible(array $auth) {
 				$uid = $auth[$uidMapping];
 			}
 
+			// make sure that a valid UID is given
+			if (empty($uid)) {
+				$this->logger->error('Uid "' . $uid . '" is not a valid uid please check your attribute mapping', ['app' => $this->appName]);
+				throw new \InvalidArgumentException('No valid uid given, please check your attribute mapping. Given uid: ' . $uid);
```
nickvergessen: Throw `NoUserFoundException` instead, so it is caught in the same way,
if not update the docs as well

schiessle: I chosed the "InvalidArgumentException" because it is used here as well if no valid user is given: https://github.com/nextcloud/user_saml/pull/174/files#diff-08ab8d6da860d4cc7a5530b6414b0f6eR238 But beside that I don't care that much

blizzz: The calles of this method wrap it into a try block and only catch NoUserFoundException however.

schiessle: yes, because for this exception we have a dedicated error page. For the other case not. As in this case the useful information are in the log I think the generic error page which points you to the log makes much more sense than a "user not provisioned" error page because that's not the problem here.

blizzz: OK

MorrisJobke:
@blizzz @schiessle Just to let you know: this went to master making it 14 only - should it be in 13? Then backport it to stable13

schiessle:
@MorrisJobke Thanks for the notice. I will backport it because I want to have it available for Nextcloud 13 users. But user_saml get's distributed in the app store, so it should be independent from the server release.

MorrisJobke:
> But user_saml get's distributed in the app store, so it should be independent from the server release.

Correct 👍

PR #183, opened 2018-02-13
Add a URL to logout the user

nickvergessen:
Enable `Enable unauthenticated endpoint to logout a user` in the admin UI

Call: `curl -X POST 'https://localhost/index.php/apps/user_saml/saml/signOut' --data 'uid=admin&secret=YourSecret'` and it will kill all temporary sessions, retaining the sync client tokens.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -44,6 +44,11 @@ return [
 			'verb' => 'GET',
 		],
 		[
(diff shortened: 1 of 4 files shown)


MorrisJobke:
@nickvergessen Conflicts

@schiessle Reviews would be nice ;)

nickvergessen:
Closing since this is going into the wrong direction.

PR #185, opened 2018-02-27
Group mapping

sergio91pt:
Improved PR #95:
- Also removes groups based on the group attribute(s).
- Supports groups with spaces (which the previous PR didn't).
- Includes unit test

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -44,7 +44,8 @@ $userBackend = new \OCA\User_SAML\UserBackend(
 	$urlGenerator,
 	\OC::$server->getSession(),
 	\OC::$server->getDatabaseConnection(),
-	\OC::$server->getUserManager()
+	\OC::$server->getUserManager(),
+	\OC::$server->getGroupManager()
 );
 $userBackend->registerBackends(\OC::$server->getUserManager()->getBackends());
 OC_User::useBackend($userBackend);
--- a/lib/Settings/Admin.php
+++ b/lib/Settings/Admin.php
@@ -112,6 +112,11 @@ class Admin implements ISettings {
 				'type' => 'line',
 				'required' => false,
 			],
+			'group_mapping' => [
+				'text' => $this->l10n->t('Attribute to map the users groups to.'),
+				'type' => 'line',
+				'required' => true,
+			],
 		];
 
 		$type = $this->config->getAppValue('user_saml', 'type');
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -421,12 +421,18 @@ class UserBackend implements IApacheBackend, UserInterface, IUserBackend {
 		self::$backends = $backends;
 	}
 
-	private function getAttributeValue($name, array $attributes) {
+	private function getAttributeKeys($name)
+	{
 		$keys = explode(' ', $this->config->getAppValue('user_saml', $name, ''));
 
-		if(count($keys) === 1 && $keys[0] === '') {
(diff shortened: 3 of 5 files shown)


schiessle:
@sergio91pt thanks for your pull request. Can you please rebase your branch on current master and resolve the merge conflicts? As soon as the merge conflicts are resolved we can review and merge it.

sergio91pt:
done

schiessle:
@sergio91pt I just tried it and it looks quite good, great work! I just wonder how to send more then one group to Nextcloud? I tried it json encoded `["group1", "group2"]` but it will always be recognized as one string.

sergio91pt:
@schiessle You need to send multiple AttributeValue's in the same SAML Attribute.

If you send the same attribute multiple times (with a different value), the SAML library throws an exception.

I tested it with Keycloak. I had to enable the "Single Group/Role attribute" in the mapper settings, because the default behavior is to send the attribute multiple times.

```xml
<saml:Attribute Name="groups" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:basic">
    <saml:AttributeValue
        xmlns:xs="http://www.w3.org/2001/XMLSchema"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="xs:string">Employees
    </saml:AttributeValue>
    <saml:AttributeValue
        xmlns:xs="http://www.w3.org/2001/XMLSchema"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="xs:string">Developers
    </saml:AttributeValue>
</saml:Attribute>
```

schiessle:
@sergio91pt thanks... I currently testing with onelogin and couldn't find a way to send a groups array... @rullzer do you have a idea how to do it with onelogin?

rullzer:
I'll check this out on thursday.

@schiessle I think you can do it with roles in onelogin. But I'm not 100% sure.

rullzer:
The code makes sense to me but I can't get onelogin to test this...

rullzer:
@sergio91pt would it be possible for you to extend the integration tests at https://github.com/nextcloud/user_saml/blob/master/tests/integration/features/Shibboleth.feature so trhat we know it works?

sergio91pt:
@rullzer I took a look and it seems I'd have to edit shiboboleth's and the ldap configuration which I'm not familiar with, so it might be complicated.

@schiessle @rullzer I think I found it. On OneLogin, use the "SAML Test Connector (IdP w/attr. incl group transform)" app. On the parameters tab, edit the Member Of field, so the "default if no values selected" is user roles.

Then if you give a user multiple roles it will send them using multiple AttributeValue's (attribute name: memberOf). 

I couldn't get my test instance (unmodified nextcloud docker image) to provision accounts with onelogin, but then again, it's the first time I test with onelogin, so I probably have it mis-configured.

schiessle:
@sergio91pt Thanks for the hint, I could test it with the "Roles". As you said I have to add multiple roles to the user, then in the attributes for the "Roles" parameters I have to set "Users Roles" as the default and the value type to "Semicolon Delimited Input (Multi-value output)"

Then in the Nextcloud settings it is important to map the groups to the "roles" parameter (lower "r", even if it is named "Roles" in oneLogin)

schiessle:
I will merge it now. Do you want to create a backport to the stable13 branch? Then it will be part of the next release.

sergio91pt:
@schiessle Done, cherry-picked without conflicts. See #205

PR #190, opened 2018-03-13
detect disabled user and show a apropriated error message

schiessle:
detect disabled user and show a appropriated error message

Steps to test:

1. setup Nextcloud with SAML
2. Disable a saml user in the user management
3. Try to login with the disabled user

Until now: it result in a redirect loop between Nextcloud and the idp

With this PR: a appropriated error message is shown.


schiessle:
any idea why the info.xml should not be correct? Didn't changed anything here...https://drone.nextcloud.com/nextcloud/user_saml/330/34 cc @nickvergessen

schiessle:
@MorrisJobke maybe you have a idea regarding the integration tests. We run them with PHP5.6 which of course result in an error on master as the min PHP version is here 7.0. I already set the min Nextcloud version here to 14, as you suggested. But the result is the same. As you can see here https://github.com/nextcloud/user_saml/blob/master/.drone.yml#L244 we are using php56 for master. I tried to change it to "rh-php70" but it doesn't work. Any idea?

schiessle:
I tried to create here a new Docker: https://github.com/nextcloud/docker-ci/pull/85 for the integration tests based on PHP7

PR #191, opened 2018-03-14
Detect disabled user stable13

schiessle:
backport of https://github.com/nextcloud/user_saml/pull/190

detect disabled user and show a appropriated error message

Steps to test:

1. setup Nextcloud with SAML
2. Disable a saml user in the user management
3. Try to login with the disabled user

Until now: it result in a redirect loop between Nextcloud and the idp

With this PR: a appropriated error message is shown.

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -29,6 +29,7 @@ if(OC::$CLI) {
 }
 
 $urlGenerator = \OC::$server->getURLGenerator();
+$l = \OC::$server->getL10N('user_saml');
 $config = \OC::$server->getConfig();
 $request = \OC::$server->getRequest();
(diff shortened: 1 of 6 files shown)


appinfo/app.php
```diff
@@ -71,6 +72,22 @@
 }
 
 $redirectSituation = false;
+
+$user = $userSession->getUser();
```
blizzz: this block could go to Appinfo\Application :) since there's enough stuff lying around here it can be moved around with a subsequent PR

schiessle: for now I just want to get it fixed and don't start a general cleanup round. We need to restructure this anyway completely from scratch once we allow multiple idp's and user back-ends

PR #193, opened 2018-03-14
Add support for mapping the quota

rullzer:
Allows to specify a customer mapping for the quota element.

I need to think about the best way to test this.

### DIFF
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -455,6 +455,11 @@ class UserBackend implements IApacheBackend, UserInterface, IUserBackend {
 		} catch (\InvalidArgumentException $e) {
 			$newDisplayname = null;
 		}
+		try {
+			$newQuota = $this->getAttributeValue('saml-attribute-mapping-quota_mapping', $attributes);
+		} catch (\InvalidArgumentException $e) {
+			$newQuota = null;
+		}
 
 		if ($user !== null) {
 			$currentEmail = (string)$user->getEMailAddress();
@@ -473,6 +478,10 @@ class UserBackend implements IApacheBackend, UserInterface, IUserBackend {
 					]
 				);
 				$this->setDisplayName($uid, $newDisplayname);
+			}
+
+			if ($newQuota !== null) {
+				$user->setQuota($newQuota);
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


lib/UserBackend.php
```diff
@@ -474,6 +479,10 @@ public function updateAttributes($uid,
 				);
 				$this->setDisplayName($uid, $newDisplayname);
 			}
+
+			if ($newQuota !== null) {
+				$user->setQuota($newQuota);
```
schiessle: should we do some additional sanity checks here, e.g. if it is an integer and >0? Or does setQuota() checks it already?

rullzer: Set quota does all kind of magic parsing. Like if it is an int it is in bytes. But you can also pass 100MB etc.

As far as I can tell we also just pass it on in the LDAP code.

lib/UserBackend.php
```diff
@@ -455,6 +455,11 @@ public function updateAttributes($uid,
 		} catch (\InvalidArgumentException $e) {
 			$newDisplayname = null;
 		}
+		try {
+			$newQuota = $this->getAttributeValue('saml-attribute-mapping-quota_mapping', $attributes);
```
schiessle: If the parameter we are looking for doesn't exists, this function returns '' (empty string). setQuota() will translate it in ' B'. I think this doesn't make sense. We should translate a empty string either to `null` so that nothing happens,  'default' or 'none'. Not sure what's the best in this case, but everything is probably better than setting the quota to ' B' if the parameter doesn't exists

rullzer: yes agreed. I would vote if none is returned we set it to 'default'. as that makes the most sense in my opinion.

rullzer:
@schiessle to test with onelogin.

1. Add a new customer user field `quota`
2. Go to your app on onelogin
3. Go to parameters
4. Add a new parameter give it a name (mine was `quota`) save it
5. edit your newly added parameter and set its value to `quota` as well (your customer attribute).
6. check include in SAML assertion
7. Now your quota attribute should be returned.

lib/UserBackend.php
schiessle: yes, this was also my initial thought... so let's go with 'default'

schiessle:
btw, tests need to be adjusted

rullzer:
Thanks for the fix @schiessle! Lets do this.

PR #196, opened 2018-03-16
Multiple user back ends

schiessle:
add option to enable multiple user back-ends in parallel (see second check box from the saml admin settings) and to give the identity provider a custom display name (see first input field):

sso2

The user will then see this landing page and can decide between SSO and "direct login" which can be any user management which authenticates directly at Nextcloud (local user management, ldap,...). The SSO button has either the default name, as show here or the display name configured above.

sso2

@rullzer maybe you can give it a try and review it? Thanks!

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -373,7 +373,7 @@ class SAMLController extends Controller {
 	}
 
 	/**
-	 * get SSO URL
+	 * get Nextcloud login URL
 	 *
 	 * @return string
 	 */
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -309,4 +309,57 @@ class SAMLController extends Controller {
 		}
 		return new Http\TemplateResponse($this->appName, 'error', ['message' => $message], 'guest');
 	}
+
+	/**
+	 * @PublicPage
+	 * @NoCSRFRequired
+	 * @OnlyUnauthenticatedUsers
+	 * @param string $redirectUrl
+	 * @return Http\TemplateResponse
+	 */
+	public function selectUserBackEnd($redirectUrl) {
+		$loginUrls = [
+			'directLogin' => $this->getDirectLoginUrl(),
+			'ssoLogin' => $this->getSSOUrl($redirectUrl)
+		];
+		return new Http\TemplateResponse($this->appName, 'selectUserBackEnd', $loginUrls, 'guest');
+	}
+
+	/**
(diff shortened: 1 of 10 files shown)


MorrisJobke:
I really like the idea. The only thing I see with this approach is that people need to know the difference somehow. Either customizable button labels (to suite the terms that are used for those login types in that environment) or some better generic terms (external and internal users - guests and members- etc) could help here to understand the difference from a nontechnical user point of view.

Maybe somebody from @nextcloud/designers has a good idea

schiessle:
@MorrisJobke in a next step I want to introduce the possibility to have multiple SSO idp's then I want to add the possibility to give them custom names (and maybe logos). But I could already here allow to rename the saml button... let me check

schiessle:
@MorrisJobke the button label is now configurable

templates/selectUserBackEnd.php
```diff
@@ -0,0 +1,20 @@
+<?php
+style('user_saml', 'selectUserBackEnd');
+
+/** @var array $_ */
+/** @var $l \OCP\IL10N */
+?>
+
+<div id="saml-select-user-back-end">
+
+<h1>Chose log in option:</h1>
```
blizzz: typo

templates/selectUserBackEnd.php
```diff
@@ -0,0 +1,20 @@
+<?php
+style('user_saml', 'selectUserBackEnd');
+
+/** @var array $_ */
+/** @var $l \OCP\IL10N */
+?>
+
+<div id="saml-select-user-back-end">
+
+<h1>Chose login option:</h1>
```
blizzz: typo is still here :) s/chose/choose

rullzer: Ah sorry missed that...

schiessle: no problem... will fix the typo quickly

lib/Controller/SAMLController.php
```diff
@@ -309,4 +309,77 @@ public function genericError($message) {
...
+	 */
+	private function getSSOUrl($redirectUrl) {
+
+		$originalUrl = '';
+		if(!empty($redirectUrl)) {
+			$originalUrl = $this->urlGenerator->getAbsoluteURL($redirectUrl);
+		}
+
+
+		$csrfToken = \OC::$server->getCsrfTokenManager()->getToken();
```
MorrisJobke: Should be injected as dependency in the constructor.

PR #206, opened 2018-04-09
WiP: Implement mapping of user's home directory

### LINKED ISSUE #202: Add support for mapping user home
User backends may provide custom home directories for users (through `getHome()`), so they don't need to be stored under `datadirectory/$uid`. The `user_ldap` app already supports this by mapping an LDAP attribute to the user's Nextcloud home.

It would be nice if custom user home locations could also be implemented in the `user_saml` app through an _optionally_ mapped attribute.

### Implementation

I already have an idea how this could be implemented in a backwards-compatible way:

1. Add a `home_directory` column to the `user_saml_users` table
2. Implement mapping from variable (SAML or env variable) to `home_directory` and store it in database
3. Implement `\OCA\User_SAML\UserBackend->getHome($uid)` with the following logic:
   -  If value of `home_directory` is empty, return `false` (which means we should fall back to the default location for backwards compatibility)
   - If the value is set, then return it (obviously)

I can try to come up with a PR for this, but only if there is some feedback available if this is feasible or not.

klada:
This is an initial proof-of-concept for mapping user homes.

Closes #202 

Signed-off-by: Daniel Klaffenbach <daniel.klaffenbach@hrz.tu-chemnitz.de>

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -24,6 +24,15 @@
 				<notnull>true</notnull>
 				<length>255</length>
 			</field>
+
+			<field>
+				<name>home</name>
+				<type>text</type>
+				<default></default>
+				<notnull>true</notnull>
+				<length>255</length>
+			</field>
+
 		</declaration>
 	</table>
 
--- a/lib/Settings/Admin.php
+++ b/lib/Settings/Admin.php
@@ -117,6 +117,12 @@ class Admin implements ISettings {
 				'type' => 'line',
 				'required' => true,
 			],
+			'home_mapping' => [
+				'text' => $this->l10n->t('Attribute to map the users home to.'),
+				'type' => 'line',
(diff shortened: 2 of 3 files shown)


schiessle:
Great! :+1: 
What is missing to move this from or PoC to something which is ready to review and merge? Please let us know if you need any help.

schiessle:
To trigger a DB upgrade to add the new column for people who have the app already in use, you need to increase the version number here: https://github.com/nextcloud/user_saml/blob/master/appinfo/info.xml#L19

klada:
> What is missing to move this from a PoC to something which is ready to review and merge?

Just some feedback :) Thanks for the input, I'll bump the version number and will take a look into fixing the test suite. I'll get back to you shortly.

klada:
See #207

PR #207, opened 2018-04-12
Implement mapping of home directory when creating a user

### LINKED ISSUE #202: Add support for mapping user home

klada:
This implements the mapping of home directories for new users.

Right now the mapping only happens when creating the user, as **changing** an existing user's home directory cannot be considered safe at the moment.

Closes #202 
Closes #206 

Signed-off-by: Daniel Klaffenbach <daniel.klaffenbach@hrz.tu-chemnitz.de>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+
+## 2.0.1
+
+### Changed
+
+- add attribute mapping for the users home directory when creating a new user
+
+
 ## 2.0.0
 
 ### Changed
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -25,6 +25,15 @@
 				<notnull>true</notnull>
 				<length>255</length>
 			</field>
+
+			<field>
+				<name>home</name>
+				<type>text</type>
+				<default></default>
+				<notnull>true</notnull>
+				<length>255</length>
+			</field>
+
 		</declaration>
 	</table>
 
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -16,7 +16,7 @@ The following providers are supported and tested at the moment:
 	* Any other provider that authenticates using the environment variable
 
 While theoretically any other authentication provider implementing either one of those standards is compatible, we like to note that they are not part of any internal test matrix.]]></description>
-	<version>2.0.0</version>
+	<version>2.0.1</version>
 	<licence>agpl</licence>
 	<author>Lukas Reschke</author>
 	<namespace>User_SAML</namespace>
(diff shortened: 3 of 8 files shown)


klada:
@schiessle Thanks for the input, this PR is now ready. I have increased the version number and I have also fixed the tests, but for some reason the `signed-off-check` failes with this error:
```
The environment variable GITHUB_TOKEN has no proper value.
```

To me this looks like a CI issue.

klada:
@schiessle Is there anything else that is missing from my side that prevents a merge?

klada:
@schiessle OK, I'll try to look into this again (hopefully) next week.

klada:
I have rebased my work on the current master branch and did some testing with the new version. It's still working :smile:

The multiple IdP-settings page also did not cause any issues, however I have only tested this against one single IdP.

@schiessle I guess you can take a look at it now.

schiessle:
@klada awesome... :slightly_smiling_face:  I will try it out tomorrow!

schiessle:
@klada can you rebase it a last time and resolve the merge conflict... Then I would merge your PR. Thanks!

klada:
@schiessle Cool, I have rebased my commit again. It was just a newline, nothing serious.

schiessle:
OK, so let's get this in... Thanks a lot for your patience @klada

klada:
@blizzz This is correct, only full paths can be used. This is equivalent to the home directory mapping provided by `user_ldap`.

blizzz:
user_ldap checks the value that was read and if it does not start with a slash it prepends the configured data directory.

schiessle:
> user_ldap checks the value that was read and if it does not start with a slash it prepends the configured data directory.

might be a nice addition, @klada could you do a follow-up PR to add this behavior here as well?

klada:
Yes, I'll look into this. Sounds like a good idea to make sure that the path is resolvable.

PR #210, opened 2018-04-25
make SAML app ready for Global Scale

schiessle:
don't auto provision users on the Global Site Selector in a Global Scale setup

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -22,6 +22,7 @@
 
 namespace OCA\User_SAML\Controller;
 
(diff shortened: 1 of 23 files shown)


lib/Controller/SAMLController.php
```diff
@@ -98,6 +99,15 @@ public function __construct($appName,
 	 * @throws NoUserFoundException
 	 */
 	private function autoprovisionIfPossible(array $auth) {
+
+
+
+
```
nickvergessen: :eyes:

Issue #172, opened 2018-01-12
New users got no default (skeleton) documents


apw1388:
We enabled the user_saml module to authenticate new users with our IdP. New users have an empty home folder within unity and the default/skeleton documents like Nextcloud manual are not stored in their homes. If we create a new local user, those documents are present. This issue happened to all users, with user_saml login, across different browsers and OS.

I don't know if this is a problem of user_saml module or Nextcloud server, but it appears if new users login by user_saml module.

### Steps to reproduce
1. Install Nextcloud 12.0.4
2. Enable user_saml module
3. Sign in with SAML IdP as a new user

### Expected behaviour
The sync clients pop up for new user is shown and default (skeleton) documents are stored in home folder. ~ 7MB of user space are used by this documents.

### Actual behaviour
The pop up is shown and no files are stored/shown in the home folder.  No space is used.

### Server configuration
**Operating system**:
CentOS 7.3
**Web server:**
Apache
**Database:**
MariaDB  5.5.52
**PHP version:**
7.0.25
**Nextcloud version:** (see Nextcloud admin page)
12.0.4
**Where did you install Nextcloud from:**
tar.bz2 file from Nextcloud homepage

**List of activated apps:**
  - activity: 2.5.2
  - b2sharebridge: 0.2.0
  - bruteforcesettings: 1.0.3
  - calendar: 1.5.7
  - comments: 1.2.0
  - contacts: 2.0.1
  - dav: 1.3.0
  - federatedfilesharing: 1.2.0
  - federation: 1.2.0
  - files: 1.7.2
  - files_pdfviewer: 1.1.1
  - files_sharing: 1.4.0
  - files_texteditor: 2.4.1
  - files_trashbin: 1.2.0
  - files_versions: 1.5.0
  - files_videoplayer: 1.1.0
  - firstrunwizard: 2.1
  - logreader: 2.0.0
  - lookup_server_connector: 1.0.0
  - nextcloud_announcements: 1.1
  - notifications: 2.0.0
  - oauth2: 1.0.5
  - password_policy: 1.2.2
  - provisioning_api: 1.2.0
  - serverinfo: 1.2.0
  - sharebymail: 1.2.0
  - survey_client: 1.0.0
  - systemtags: 1.2.0
  - theming: 1.3.0
  - twofactor_backupcodes: 1.1.1
  - updatenotification: 1.2.0
  - user_ldap: 1.2.1
  - user_saml: 1.4.0
  - workflowengine: 1.2.0

[...]

PR #220, opened 2018-06-18
Create skeleton and dispatch first login event for auto-created accounts

FGIKCM:
Code taken from 'regular' login method do create skeleton and dispatch event of the user creation.
A better idea would be to directly use the `prepareUserLogin()` method of `lib/private/User/Session.php`, but as it is private...

This should fix https://github.com/nextcloud/user_saml/issues/172


schiessle:
@FGIKCM thanks for your fix... I will have a look

schiessle:
I also don't see a easy way to do this in a more generic way at the moment... so let's do it.

Issue #223, opened 2018-06-26
Allow authentication for specific implementations of Public Key Infrastructure


fgsl:
### Overview
Aim of this change is allowing specific implementations for authentication based on security tokens.
Nowadays, user_saml app accepts two types of login: saml and environment-variable. But there are different sets of OIDs for different public key infrastructures. Then, there are different strategies of authentication.
The proposal of implementation is allowing that programmers use properly strategy for public infrastructure public they are working with.
It's no intention to discard current implementation, but to allow that it can be adapted according reality of environment.

### Steps to reproduce
1. Extends OCA\User_SAML\Strategies\AuthenticationAbstractPKIStrategy.
2. Extends OCA\User_SAML\UserBackend.

### Expected behaviour
A user logs in Nextcloud using a security token.

### Actual behaviour
Nextcloud does not support login for any security token.

### Server configuration
It is necessary to config certificate in web server. Below are items that must to be set in apache site configuration file.
`
<VirtualHost *:443>

    UseCanonicalName On
    ServerName [SERVERNAME]

    SSLEngine on
    SSLProxyEngine on

    SSLCertificateFile      [CRT FILE]
    SSLCertificateKeyFile   [KEY FILE]
    SSLCACertificateFile    [CRT FILE]
    SSLVerifyDepth 3
    SSLOptions +ExportCertData +StdEnvVars

    # Initialize the special headers to a blank value to avoid http header forgeries
    RequestHeader set SSL_CLIENT_CERT ""
    RequestHeader set SSL_CLIENT_VERIFY ""

    SSLVerifyClient none
    <Location "/index.php/apps/user_saml/saml/login">
        RewriteEngine on
        RewriteCond %{HTTPS} off
        RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [L]

        SSLVerifyClient optional
        SSLVerifyDepth 3
        SSLOptions +ExportCertData +StdEnvVars

        RewriteCond %{SSL:SSL_CLIENT_VERIFY} !^SUCCESS$
        RewriteRule .* /apps/user_saml/ssl_error.html [R=302,L]
        RequestHeader set SSL_CLIENT_CERT "%{SSL_CLIENT_CERT}s"
        RequestHeader set SSL_CLIENT_VERIFY "%{SSL_CLIENT_VERIFY}s"
    </Location>

    SSLProtocol All -SSLv2 -SSLv3

    SSLHonorCipherOrder  on

[...]

PR #224, opened 2018-06-26
Allow authentication for specific implementations of PKI #223

fgsl:


blizzz:
I am going through the open PRs and a there is a bunch of dusty ones here 😅

Is this one still relevant to you @fgsl ?

blizzz:
Alright, thank you for the feedback @fgsl! 

I close it hence, could be re-used in case something like this pops up at some point again.

PR #230, opened 2018-07-11
Follow the redirect url on direct login

rullzer:
This makes sure the auth flow also works with the direct login.


schiessle:
@rullzer please have a look at the failing integration tests

rullzer:
Mmm it looks like the intergration test fail because of LDAP stuff... I can look at that later

schiessle:
@rullzer... I just restarted the tests and now they passed. Think it was just a ci hiccup

PR #236, opened 2018-08-13
update to php-saml 3.0.0

schiessle:
get rid of the mcrypt dependency. Seems we are still running in this error: https://github.com/onelogin/php-saml/issues/249 ... But let's try to find out more


schiessle:
@rullzer I updated the saml_php lib to 3.0.0dev to get rid of the mcrypt dependency and everything seems to work. I'm not really happy with creating a fake root to build the correct base url, but I couldn't find a better way: https://github.com/nextcloud/user_saml/pull/236/commits/6b8bebc252aebccf90814b963addff351269ae72#diff-0e033fd8797499a87a6bf2ffccef5c3cR31

rullzer:
@schiessle it seems you didn't add all the new files?

schiessle:
@rullzer ok, now it looks better. :wink:

schiessle:
@rullzer now with php-saml 3.0.0 stable! :tada:

rullzer:
After this is in I'll enable dependabot so we get automatic upgrades

PR #237, opened 2018-08-14
Add reset option

schiessle:
add a button to reset the config and go back to the initial setup to start over.

To lower the risk that the user will lose the configuration by accident we only remove the "saml type", this is enough to show again the default login screen and let you start the configuration from scratch. But you will stiff have the previous configuration if you chose one of the SSO type


blizzz:
what happens with the old config  once a new one is being started?

schiessle:
if you chose either "SAML" or "environment variables" the old config will be shown again and then you can either adjust it or delete the tab and start over from scratch.

PR #240, opened 2018-08-17
Add orignalURL to redirect when using SSO

jbpin:
When using environment variable setting the originalUrl params is not used for redirecting the request.
This request is to add originalUrl as a redirect value when using SSO with environment variables or to fallback to the root path.

PR #241, opened 2018-08-17
Add orignalURL to redirect when using SSO

schiessle:
rebased from here: https://github.com/nextcloud/user_saml/pull/240


jbpin:
🤝thank you.

PR #247, opened 2018-09-05
Fix typo in german translation

winpat:
Fix typo in german translation.


schiessle:
Thanks for spotting the typo! Translation are done at Transifex: https://www.transifex.com/nextcloud/nextcloud/ and then copied over here regularly. I added the fix to Transifex now. I would be happy to welcome you on Transifex in our translator team! :slightly_smiling_face:

PR #251, opened 2018-10-04
add missing use clause

schiessle:
an easy one... @nickvergessen do you remember? :wink:


nickvergessen:
@morrisjobke is saml shipped? If fso this should be in 14.0.2

rullzer:
@nickvergessen nope appstore

PR #260, opened 2018-11-09
add a combobox instead of buttons to select the login method if more then 4 different IDPs are configured

schiessle:
If you have multipe IDPs the login page looks like this:

image

This looks nice and is intuitive to use but it doesn't scale for a large number of IDPs, so I made this changes to switch to a combobox if there are more then 4 IDPs configured, in this case it looks like this:

image

### DIFF
--- a/js/selectUserBackEnd.js
+++ b/js/selectUserBackEnd.js
@@ -1,0 +1,8 @@
+$(window).load(function() {
+
+	$(document).on('click', 'option', function() {
+		var target = $(this).val();
+		window.location.href = target;
+	});
+
+});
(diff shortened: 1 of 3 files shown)


rullzer:
Just some nitpicking. But I would assume that 'direct login' is kind of the edge case in 90% of the cases right ;) So maybe just move that to the bottom?

schiessle:
@rullzer good point, moved the direct login down

schiessle:
Did another small improvements, I don't pre-select a IDP but have a descriptive place holder because otherwise it might confuse the user like "the right idp is already selected, where is the submit button"

image

PR #263, opened 2018-11-10
Adding setting to whitelist groups that are protected from being removed

Deltachaos:
We have the usecase that some users should be enhanced with privileges that are not part of the SAML response.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -33,6 +33,7 @@ While theoretically any other authentication provider implementing either one of
 	<screenshot>https://raw.githubusercontent.com/nextcloud/user_saml/master/screenshots/1.png</screenshot>
 	<screenshot>https://raw.githubusercontent.com/nextcloud/user_saml/master/screenshots/2.png</screenshot>
 	<dependencies>
+		<lib>mcrypt</lib>
 		<nextcloud min-version="14" max-version="15" />
 	</dependencies>
 	<settings>
--- a/lib/Settings/Admin.php
+++ b/lib/Settings/Admin.php
@@ -120,6 +120,11 @@ class Admin implements ISettings {
 				'type' => 'line',
 				'required' => true,
 			],
+			'protected_group_mapping' => [
+				'text' => $this->l10n->t('Protected Groups'),
+				'type' => 'line',
+				'required' => true,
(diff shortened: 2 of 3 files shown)


schiessle:
@Deltachaos thanks for your pull request.

Do you use this "protected groups" also in some other scenarios or do you add it to both Nextcloud and your IdP as a special workaround for Nextcloud? Do you know other SAML connectors which offer a similar configuration option? I'm asking because while I see the rationals behind your PR, it looks like a quite specific use case so I wonder if we should add settings for such highly specific stuff.

Maybe a "remove unused groups" checkbox next to the group mapping setting would be a more generic and general useful solution? Not as customizable as your approach because it would be a "all or nothing" setting, but maybe easier to understand, you don't need a additional parameter from the IDP and more generic?

What other people think... cc @rullzer

Deltachaos:
I have only one installation of nextcloud so therefore i do not use it in other installations yet. The problem with a checkbox "remove unused groups" would be that if a group that was created by the SAML login for a user would not be revoked on login anymore if it is not part of the SAML login anymore.

I think that it is a quite common feature to have some groups that you don't control centralized because they are very application specific.

schiessle:
I thought a little bit more about it.

> We have the usecase that some users should be enhanced with privileges that are not part of the SAML response.

this means that we are talking about groups not created by the SAML response but by the Nextcloud admin in the user management, right? SO with your proposal the admin would always need to adjust the parameter stored in the IDP after creating a local group. In case the Nextcloud admin is not the same as the IDP admin he would always have to ask his colleague to update the IDP parameter.

How about we track instead at Nextcloud which group was created my a SAML response and only remove the user from this groups if they are no longer listed in the SAML response, but not from groups created by Nextcloud? This way we could manage the list of "protected groups" automatically. Would this solve the problem?

Deltachaos:
This would be a more elegant solution, yes.

schiessle:
@Deltachaos would you be interested in implementing it?

Deltachaos:
@schiessle in general yes. But i am not so deep in the nextcloud code. Could you give me a hint where is should put this information?

Deltachaos:
Closed in favor of #311

PR #273, opened 2018-11-22
fix select idp combobox so that it also works with chromium

schiessle:
fix select idp combobox so that it also works with chromium


schiessle:
Just tested it with Firefox on Android and there it still doesn't work...

Any idea why?

PR #276, opened 2018-11-23
Update version on master

nickvergessen:

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -30,7 +30,7 @@ While theoretically any other authentication provider implementing either one of
 	<screenshot>https://raw.githubusercontent.com/nextcloud/user_saml/master/screenshots/1.png</screenshot>
 	<screenshot>https://raw.githubusercontent.com/nextcloud/user_saml/master/screenshots/2.png</screenshot>
 	<dependencies>
-		<nextcloud min-version="16" max-version="16" />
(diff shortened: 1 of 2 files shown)


schiessle:
as said in the other issue, let's keep it 14 to 16. I will always create a branch if we introduce incompatibilities to the server, otherwise the app can be decoupled from the server version.

MorrisJobke:
integration tests fail

appinfo/info.xml
```diff
@@ -16,15 +16,15 @@ The following providers are supported and tested at the moment:
...
+	<version>2.1.0</version>
 	<licence>agpl</licence>
 	<author>Lukas Reschke</author>
 	<namespace>User_SAML</namespace>
 	<types>
 		<authentication/>
 	</types>
 	<documentation>
-		<admin>https://docs.nextcloud.com/server/13/admin_manual/configuration_server/sso_configuration.html</admin>
+		<admin>https://docs.nextcloud.com/server/16/admin_manual/configuration_server/sso_configuration.html</admin>
```
nickvergessen: @schiessle also use a go.php link here please

schiessle:
> integration tests fail

seems to be again an issue with the LDAP docker... This problem already existed a few days ago, then magically disappeared yesterday and now is back again :confused:

PR #277, opened 2018-11-23
Update info.xml

nickvergessen:
Should use a go.php link for docs @schiessle


schiessle:
I would prefer to keep min version to 14 until we create a incompatibility with older version.

nickvergessen:
there is a stable14 branch, then kill those branches please, otherwise it creates confusion :man_shrugging:

schiessle:
I'm happy to kill it... Don't use it anyway to build releases.

PR #279, opened 2018-11-26
Add sanity checks for user home directory

klada:
When the mapped user home is not a fully qualified path name we'll fall
back to setting the mapped home below the server's datadirectory. This
provides consistent behavior with the "user_ldap" app which uses the same
fallback/safety mechanism.

See #207


klada:
The code was copied and adjusted from `user_ldap`.

PR #281, opened 2018-12-03
Allow multiple user backends env variables

schiessle:
Allow to login with the local user management and with environment variables in parallel


blizzz:
do we still have that use case? There is also no logout handling for env mode (currently suppressed).

blizzz:
closing for inactivity

Issue #282, opened 2018-12-06
SAML & LDAP - mail attribute not populated


maghog:
When login using SAML the mail attribute never gets populated/updated.
Login using LDAP works as expected and the mail attribute gets populated/updated.

Is this a known limitation? It's a severe issue because the user mail address is needed in Nextcloud (share, Talk...)
 
Nextcloud 14.04
LDAP 1.4.0
SAML 2.1.0

ChristophWurst:
cc @blizzz @schiessle

Issue #268, opened 2018-11-19
Skeleton files are not deployed (Nextcloud14)


devi69:
We have set up the user_saml application to be able to authenticate users on LDAP.
The default document in the skeleton directory should be placed at the first login, but nothing was created in the user's folder.

This problem occurs when logging in via user_saml.
When an LDAP user performs a direct login at the first login, folders and files under the skeleton are created.

This problem seems to be very similar to #172 which occurred at the time of Nextcloud 12.

### Steps to reproduce
1.Install Nextcloud 14.0.3
2.Enable user_ldap app
3.Enable user_saml app
4.Sign in with SAML IdP as a new user

### Expected behaviour
For the new user, the farst run wizard is displayed and the default document placed in the skeleton directory is saved in the home folder.

### Actual behaviour
Skeleton files and directories are not created in the home folder after the first run wizard is displayed.

### Server configuration

**Operating system**:
CentOS Linux release 7.5
**Web server:**
nginx-1.14.0
**Database:**
MariaDB-10.0.36
**PHP version:**
php-7.1.23-1  
php-fpm-7.1.23-1
**Nextcloud version:** (see Nextcloud admin page)
14.0.3
**Where did you install Nextcloud from:**
tar.bz2 file from nextcloud.com

**List of activated apps:**

```
  - accessibility: 1.0.1
  - activity: 2.7.0
  - cloud_federation_api: 0.0.1
  - comments: 1.4.0
  - dav: 1.6.0
  - federatedfilesharing: 1.4.0
  - federation: 1.4.0
  - files: 1.9.0
  - files_pdfviewer: 1.3.2
  - files_sharing: 1.6.2
  - files_texteditor: 2.6.0
  - files_trashbin: 1.4.1
  - files_versions: 1.7.1
  - files_videoplayer: 1.3.0
  - firstrunwizard: 2.3.0
  - gallery: 18.1.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.2.0
  - nextcloud_announcements: 1.3.0
  - notifications: 2.2.1
  - oauth2: 1.2.1
  - password_policy: 1.4.0
  - provisioning_api: 1.4.0
  - serverinfo: 1.4.0
  - sharebymail: 1.4.0
  - support: 1.0.0
  - survey_client: 1.2.0
  - systemtags: 1.4.0
  - theming: 1.5.0
  - twofactor_backupcodes: 1.3.1
  - updatenotification: 1.4.1
[...]

PR #286, opened 2018-12-14
always create user in the SAML back-end and update the attributes

schiessle:
create user in the SAML back-end and update the attributes when the user was found on another back-end during login

fix https://github.com/nextcloud/user_saml/issues/268

@maghog @devi69 could you try to apply this two line fix to your installation and report back if this solves the problem? This might also fix https://github.com/nextcloud/user_saml/issues/282

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -144,7 +144,6 @@ class SAMLController extends Controller {
 				// help with it and make the user known
 				$this->userManager->search($uid);
 				if($this->userManager->userExists($uid)) {
-					$this->userBackend->initializeHomeDir($uid);
 					return;
 				}
 				throw new NoUserFoundException('Auto provisioning not allowed and user ' . $uid . ' does not exist');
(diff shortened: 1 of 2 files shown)


maghog:
This fix did not solve the problem...

lib/Controller/SAMLController.php
```diff
@@ -144,6 +144,8 @@ private function autoprovisionIfPossible(array $auth) {
 				// help with it and make the user known
 				$this->userManager->search($uid);
 				if($this->userManager->userExists($uid)) {
+					$this->userBackend->createUserIfNotExists($uid, $auth);
```
blizzz: my gut feeling is not positive about that line

schiessle:
@blizzz I updated the PR to only initialize the home directory of the user. Because you are right, creating the user in the saml back-end doesn't make sense if the user already exists in the ldap back-end.

What do you think?

blizzz:
@schiessle Uff… why does it not suffice having it with the user session? I am not a fan of code duplication, especially across repos… this is not really maintainable. If it is necessary, perhaps we can move it into a trait or public method on a service. But rather I prefer to keep user backends out of dealing with storage aspects. Do you know why the user session logic does not kick in?

schiessle:
@blizzz because in the session it happens in "prepareUserLogin()" which never happens if login happens on a external service.

blizzz:
@schiessle  but, does it circumvent the whole LoginController::tryLogin()?

jsalatiel:
Is https://github.com/nextcloud/user_saml/issues/268 fixed by this ?

Issue #208, opened 2018-04-16
IdP initiated SSO


gkrol-zermelo:
I'd like to use IdP initated SSO with Nextcloud. The reason for this is that we have different servers/URLs for different customers and those should all be able to access Nextcloud.

### Steps to reproduce
1. Do NOT go to Nextcloud
2. POST a SAML response to /index.php/apps/user_saml/saml/acs

### Expected behaviour
I'd like the user to be logged in to Nextcloud, whatever the state of the user previously was (not logged in, logged in, expired session).

### Problems and my fixes
Normally you'd get "null" as a response.
Fix: comment out the following lines in assertionConsumerService, so it looks like this:
```
                // if(is_null($AuthNRequestID) || $AuthNRequestID === '') {
                //       return;
                // }
```
Then, logging in works. However, when you perform the SAML login again, you get an error "The user is already logged in" (or something like that). This is fixed by commenting out:
```
// @OnlyUnauthenticatedUsers
```
The code now looks like this:
```
         /**
         * @PublicPage
         * @NoCSRFRequired
         * @UseSession
         * // @OnlyUnauthenticatedUsers
         * @NoSameSiteCookieRequired
         *
         * @return Http\RedirectResponse|void
         */
        public function assertionConsumerService() {
                $AuthNRequestID = $this->session->get('user_saml.AuthNRequestID');
                // if(is_null($AuthNRequestID) || $AuthNRequestID === '') {
                //       return;
                // }
```
This seems to work fine, however I sometimes get an internal error, with in the logs:
```
Doctrine\DBAL\Exception\UniqueConstraintViolationException: An exception occurred while executing 'INSERT INTO "oc_authtoken"("uid","login_name","name","token","type","remember","last_activity","last_check") VALUES(?,?,?,?,?,?,?,?)' with params ["<SNIP>", "<SNIP>", "<SNIP>", "<SNIP>", 0, 0, 1523868098, 1523868098]: SQLSTATE[23505]: Unique violation: 7 ERROR: duplicate key value violates unique constraint "authtoken_token_index" DETAIL: Key (token)=(<SNIP>) already exists.
```
I expect this has to do with an expired session.

### Questions

Now my actual question: is this a good way to add IdP initiated login to the user_saml plugin? Any security implications? How should I handle the expired session?

When I've got everything figured out, would you be interested in me adding a "Allow IdP initiated SSO" setting?

jagland:
[...]

PR #297, opened 2019-01-16
allow IDP to initiate login

schiessle:
if the user starts the login process on the IDP, we always assume that
the first configured IDP is used (this doesn't work for multiple IDP's
for now) and tries to login the user.

Steps to test:

- Setup an IDP, e.g. onelogin
- Go directly to the IDP (not to the Nextcloud)
- Select there the "Nextcloud app"
- You will be redirected to Nextcloud

-> Current behavior: You will see an error page
-> with this PR: you should be logged in


nbitouze:
Thank you for this PR, we would be interested in this!

There is a related discussion at #208. It is suggested that the `@OnlyUnauthenticatedUsers` annotation be removed from `assertionConsumerService()` to avoid getting a "User is already logged in" error.

I am also wondering whether the SAML best practice for the `$this->session->set('user_saml.OriginalUrl', $this->request->getParam('originalUrl', ''));` functionality would be to use a request parameter (as in the current pull request) or an SAML attribute (like `$auth->getAttribute('originalUrl')`.

kvn:
@blizzz can we help to get this reviewed and merged? IdP initiated login is currently broken. Many thanks!

mateuszdrab:
Just like @kvn, I'd like to add myself to this request.

lib/Controller/SAMLController.php
```diff
@@ -243,11 +243,17 @@ public function getMetadata($idp) {
 	public function assertionConsumerService() {
 		$AuthNRequestID = $this->session->get('user_saml.AuthNRequestID');
 		$idp = $this->session->get('user_saml.Idp');
+		$idp = $idp === null ? 1 : $idp;
```
blizzz, line 246:
```suggestion
		$idp = $this->session->get('user_saml.Idp') ?: 1;
```

lib/Controller/SAMLController.php
```diff
@@ -243,11 +243,17 @@ public function getMetadata($idp) {
...
+		$auth = new Auth($this->SAMLSettings->getOneLoginSettingsArray($idp));
+
 		if(is_null($AuthNRequestID) || $AuthNRequestID === '' || is_null($idp)) {
```
blizzz, line 249: `is_null($idp)` will now never be true, can be removed

lib/Controller/SAMLController.php
blizzz: I would not assume that 1 is always a valid id.

blizzz (commented):
hard coded ID  1 looks suspicious to me, otherwise :lipstick:

Issue #293, opened 2019-01-08
Unable to map multiple groups to a user.


pyhrr0:
### Steps to reproduce
1. Set-up Nextcloud 15.
1. Install/Enable the "SSO & SAML authentication" app.
3. Configure the app to use the following environment variables: `REMOTE_USER`, `REMOTE_USER_GROUPS`.
4. Configure your webserver to set the variables listed above.  
(e.g. REMOTE_USER=foo and REMOTE_USER_GROUPS=bar,baz)

### Expected behaviour
Since the placeholder of user_saml's group-mapping option contains the following: `Attribute to map the users groups to`, I'd expect multiple groups to be supported.

### Actual behaviour
Unfortunately the value inside of REMOTE_USER_GROUPS is treated as a literal string, and thus it's only possible to map a single group to a user. 

### Server configuration
**Operating system**:
Debian 9.6

**Web server:**
Nginx 1.10.3

**Database:**
SQLite 2.8.17

**PHP version:**
PHP 7.0

**Nextcloud version:** (see Nextcloud admin page)
15.0.0

**Where did you install Nextcloud from:**
A tarball.

**List of activated apps:**
  - accessibility: 1.1.0
  - activity: 2.8.2
  - cloud_federation_api: 0.1.0
  - comments: 1.5.0
  - dav: 1.8.0
  - federatedfilesharing: 1.5.0
  - federation: 1.5.0
  - files: 1.10.0
  - files_pdfviewer: 1.4.0
  - files_sharing: 1.7.0
  - files_texteditor: 2.7.0
  - files_trashbin: 1.5.0
  - files_versions: 1.8.0
  - files_videoplayer: 1.4.0
  - firstrunwizard: 2.4.0
  - gallery: 18.2.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.3.0
  - nextcloud_announcements: 1.4.0
  - notifications: 2.3.0
  - oauth2: 1.3.0
  - password_policy: 1.5.0
  - provisioning_api: 1.5.0
  - serverinfo: 1.5.0
  - sharebymail: 1.5.0
  - support: 1.0.0
  - survey_client: 1.3.0
  - systemtags: 1.5.0
  - theming: 1.6.0
  - twofactor_backupcodes: 1.4.1
  - updatenotification: 1.5.0
  - user_saml: 2.1.0
  - workflowengine: 1.5.0

**Nextcloud configuration:**
```
{
    "system": {
[...]

PR #307, opened 2019-02-16
Create and merge array from (multi-valued) attributes

mrvanes:
When consuming attributes from headers instead of env, multi-valued attributes are sometimes header name postfixed (_01, _02) or merged using a separator. The former can not easily be solved without extra logic, but the latter can be exploded using the separator character. The advantage is that everything will now be cast into an array, so that values can always be array_merged. Even multi separator-merged-multi-valued attributes will result in the correct multi-valued value this way.

This will solve https://github.com/nextcloud/user_saml/issues/293

TODO: I lack the time and Nextcloud app development experience to introduce a separator setting in UI.
Please add one before merging this PR.
Just make sure to replace an empty separator "" with "\0" so that explode still works and strings aren't unexpectedly cut into smaller pieces. \0 should not naturally occur in strings.

### DIFF
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -586,19 +586,8 @@ class UserBackend implements IApacheBackend, UserInterface, IUserBackend {
 		$value = '';
 		foreach($keys as $key) {
 			if (isset($attributes[$key])) {
-				if (is_array($attributes[$key])) {
-					foreach ($attributes[$key] as $attribute_part_value) {
-						if($value !== '') {
-							$value .= ' ';
-						}
-						$value .= $attribute_part_value;
-					}
-				} else {
-					if($value !== '') {
-						$value .= ' ';
-					}
-					$value .= $attributes[$key];
-				}
+				$array = explode(";", $attributes[$key]);
+				$value = array_merge($value, array_values($array));
 			}
 		}
 


tuxcrafter:
What lines can I add to the code to print some debug information.

My saml is passing an array of groups, but I only get one group in my user preferences?

```
[Tue Mar 12 23:45:11.871725 2019] [:error] [pid 3848] [12/Mar/2019:23:45:11]  DEBUG(ipsilon/providers/common.py:93 Redirect.debug()): saml2: jelledj's attributes: {'_groups': ['generalinformation', 'communications'], 'surname': 'de Jong', '_extras': {'sssd': {}}, 'groups': ['generalinformation', 'communications'], '_auth_type': 'password', 'givenname': 'Jelle', 'email': 'jelledejong@somewhere.nl'}
[Tue Mar 12 23:45:11.873282 2019] [:error] [pid 3848] [12/Mar/2019:23:45:11]  DEBUG(ipsilon/providers/common.py:93 Redirect.debug()): saml2: value de Jong
[Tue Mar 12 23:45:11.874799 2019] [:error] [pid 3848] [12/Mar/2019:23:45:11]  DEBUG(ipsilon/providers/common.py:93 Redirect.debug()): saml2: value generalinformation
[Tue Mar 12 23:45:11.876352 2019] [:error] [pid 3848] [12/Mar/2019:23:45:11]  DEBUG(ipsilon/providers/common.py:93 Redirect.debug()): saml2: value communications
[Tue Mar 12 23:45:11.877856 2019] [:error] [pid 3848] [12/Mar/2019:23:45:11]  DEBUG(ipsilon/providers/common.py:93 Redirect.debug()): saml2: value Jelle
[Tue Mar 12 23:45:11.879394 2019] [:error] [pid 3848] [12/Mar/2019:23:45:11]  DEBUG(ipsilon/providers/common.py:93 Redirect.debug()): saml2: value jelledejong@somewhere.nl
```

blizzz (changes requested):
I am going through the open PRs and a there is a bunch of dusty ones here 😅

Now the behaviour is quite different though and no callees of that method are adjusted. When dealing with different formats we should also have a set of unit tests for them.

mrvanes:
This is so old I don't even have a requirement for it anymore, but to answer the question: the behaviour never changed so that any caller was confused. The change just resulted in a more consistent representation of multi-valued attributes. Back then, the change correctly let me supply multi-valued attributes as semi-column separated strings and then profit...

blizzz:
> This is so old I don't even have a requirement for it anymore

Yeah, sorry about that. 

Maybe  I am missing some, but back then only a string was returned. But maybe that was the misbehaviour. Maybe the callees changed since; need to give it a more thorough look. 

Nevertheless, thanks for the PR and for the responsivity.

blizzz:
Replacing this with a slightly different approach in #954. The logic was present elsewhere where built-in SAML mode was not affected.

PR #311, opened 2019-02-28
Add group mapping prefix to separate SAML groups from system/other groups

anicoa:
This pr implements the discussed approach in pr #263 and isolates SAML groups from other Nextcloud groups. Now it's possible to add SAML-accounts to local groups (especially the admin-group) not getting removed after next login.

For this a new setting 'group mapping prefix' is introduced. This value is added as a prefix to all SAML-groups, group-operations within the user_saml login process are only applied to groups starting with the 'group mapping prefix'.

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -38,6 +38,22 @@ try {
 	\OC::$server->getLogger()->logException($e);
 	return;
 }
+
+\OC::$server->registerService('SAMLGroupDuplicateChecker', function() use($config) {
+    return new OCA\User_SAML\GroupDuplicateChecker(
+        $config,
+        \OC::$server->getGroupManager(),
+        \OC::$server->getLogger()
+    );
+});
+
+\OC::$server->registerService('SAMLGroupManager', function() {
+    return new OCA\User_SAML\GroupManager(
+        \OC::$server->getDatabaseConnection(),
+        \OC::$server->query('SAMLGroupDuplicateChecker')
+    );
+});
+
 $samlSettings = new \OCA\User_SAML\SAMLSettings(
 	$urlGenerator,
 	$config,
@@ -51,12 +67,15 @@ $userBackend = new \OCA\User_SAML\UserBackend(
 	\OC::$server->getSession(),
 	\OC::$server->getDatabaseConnection(),
 	\OC::$server->getUserManager(),
-	\OC::$server->getGroupManager(),
+	\OC::$server->query('SAMLGroupManager'),
 	$samlSettings,
 	\OC::$server->getLogger()
 );
 $userBackend->registerBackends(\OC::$server->getUserManager()->getBackends());
 OC_User::useBackend($userBackend);
+
+$groupBackend = new \OCA\User_SAML\GroupBackend($config, \OC::$server->query('SAMLGroupManager'));
+\OC::$server->getGroupManager()->addBackend($groupBackend);
 
 $params = [];
 
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -78,4 +78,34 @@
 			</field>
 		</declaration>
 	</table>
+
+	<table>
+		<name>*dbprefix*user_saml_group_members</name>
+		<declaration>
+			<field>
+				<name>uid</name>
+				<type>text</type>
+				<default></default>
+				<notnull>true</notnull>
+				<length>64</length>
+			</field>
+
+			<field>
+				<name>group</name>
+				<type>text</type>
+				<notnull>true</notnull>
+				<length>128</length>
+				<default></default>
+			</field>
+
+			<index>
+				<name>idx_group_members</name>
+				<unique>true</unique>
+				<field>
+					<name>group</name>
+					<name>uid</name>
+				</field>
+			</index>
+		</declaration>
+	</table>
 </database>
--- a/lib/GroupBackend.php
+++ b/lib/GroupBackend.php
@@ -1,0 +1,129 @@
+<?php
+
+namespace OCA\User_SAML;
+
+use OCP\Group\Backend\ABackend;
+use OCP\IConfig;
+
+class GroupBackend extends ABackend {
+	/**
+	 * @var IConfig
+	 */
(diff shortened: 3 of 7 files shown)


schiessle:
@anicoa Thanks a lot! The only thing I'm worrying a bit is the UX experience here. Because the user will now also see `<prefix>groupname` instead of just `groupname` because we don't have display names for groups until now.

Unfortunately I don't really have a better idea. We could create a `user_saml_group` table and track there the groups created by SAML. But I fear this will come with a huge performance penalty.

@rullzer @blizzz any ideas?

Deltachaos:
@schiessle i would suggest to may in some future introduce display names. Or may we could add another column in the groups table with a "type" information. So that we store the type "saml" for those groups and the saml plugin then should only work on groups using this type. This then would not require a prefix at all.

blizzz:
The backend support for group display names should be in place, iirc, but it is not used anywhere yet. 

Also, I don't know why groups were created inside the local groups backend. If the saml backend shall provision them, it also should take care of themselves and provide a group backend. As it does with users.

Deltachaos:
@blizzz @schiessle So let me summarize: We all agree on that SAML should manage its own groups and not interfere with groups assigned manual. We have different ideas how to do this:

1. Adding a prefix
2. Adding a prefix but show only display names
3. Adding  a type to groups
4. Creating a custom group backend

I think first two options are not really good compared to the others. Showing a display name also makes the whole prefix handing even more intransparent.

So the question is: which option of the last two should be implemented? If it is the third it increases the complexity for authentication plugins to do the job. Its probably still ok, but if i would need to implement this i would at least need a hint how a custom group backend gets implemented.

However, i would love to see having a decision on this topic before the weekend to get this feature on the road quickly. At least i know, that if the decision is made after the weekend, the feature will probably not be implemented for a long time because of very limited time resources.

blizzz:
Go with number for. I bet you can take most stuff right out of the local database group backend: https://github.com/nextcloud/server/blob/master/lib/private/Group/Database.php 

It is also the more modern approach where capabilities are divided into interfaces.

For registration, initialize and register the backend here https://github.com/nextcloud/user_saml/blob/master/appinfo/app.php#L56 . For registration, it is just `\OC::$server->getGroupManager()->addBackend($groupBackend);`

If you want another one for comparison, you can look into the LDAP's one: https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/Group_LDAP.php, it implements the old \OCP\GroupInterface though.

Deltachaos:
@schiessle @blizzz I have implemented a custom group backend now and updated this PR. The prefix option is still there, but don't need to be used. In some scenarios it is may be useful.

blizzz:
Ah, we should move away from the `database.xml` here towards miration. Out of scope for this PR though, that's fine I guess.

Formaliities: please follow the coding style as described in: https://docs.nextcloud.com/server/16/developer_manual/general/codingguidelines.html

> I have implemented a custom group backend now and updated this PR. The prefix option is still there, but don't need to be used. 

Might be troublesome when changed in production? 

With the previous solution to have groups being create as local groups, there is no way to migrate them, somehow? Perhaps as a repair job, the only criteria i can think of was if the group contains saml users only, but that's far from bullet proof.

When creating group IDs here, it is not checked whether such are already existing. Group IDs (as user ids) must be unique in the instance. This might lead to strange side effects when, e.g. local groups with the same id are already existing and a new one is created.

Deltachaos:
Hi @blizzz,

i will change the code style according to your guidelines.

About the migration: There is in my opinion no way to migrate the local groups as we dont have any information if the groups are created by the SAML plugin.

How should it behave if a local group exists? Should it ignore this group?

blizzz:
> About the migration: There is in my opinion no way to migrate the local groups as we dont have any information if the groups are created by the SAML plugin.

What about a semi-automatic runtime approach? 

* When the user logs in
* and the groups are provided
* and a local group with that name exists
* and that local group has only SAML users
* then, either
  * A) automatically convert the group 
  * B) notify the admin and provide a CLI command to migrate the group from local to saml backend

However, this would be expensive, causing a delay on login. It's likely enough to happen once per user, and this could be left to a background job. 

> How should it behave if a local group exists? Should it ignore this group?

What we do in LDAP: if a group exists  with the same name (any backend  – for LDAP we compare DN and UUID – if it's a different entry it's a name collision)  then we append a 4 digit number to create a unique group. Now with SAML I am not sure we get any other info than the group name? Then it is not an alternative. And I am not sure what is the best approach.

* use existing groups: it is some sort of group mapping, but only one-directional (creating local groups over existing other groups won't work). Therefore, inconsistent.
* use a different name: for instance by appending "(SAML)", but that's technical information end users could be irritated by and it would be inconsistent  if it only happens to duplicates. This would be more confusing. Appending an increaing number (1) (2) (..) would not help to differentiate either.
* ignore and write  a warning to the log file. Perhaps best in combination with the CLI tool mentioned ↑ so the admin can further ignore, or migrate. Sounds most predictable, consistent and least confusing.

Deltachaos:
@blizz i have implemented the logging for group conflicts now. So i guess this PR could be merged now?

Deltachaos:
@schiessle Merge conflict is resolved

schiessle:
I tested it and it works great! :+1: Also code looks good to me.

The only remaining problem are the failing tests, maybe you can have a look @anicoa or @Deltachaos ? Thanks

blizzz:
What is the state currently here?

Browsing a bit through the code I would suggest to 
* have a dedicated group table. Allows to record the group name as provided by SAML and to have an internal ID and create a mapping.
* makeing use of group display names so present it more user friendly, perhaps with (SAML) in brackets.
* the classes don't need to be made available on the global scope (it's not a biggie though)
* use the query builder instead of hand crafting sql queries

The list should not be demotivating, it's in general a big task. We may move this PR to this repo for easier collaboration. I'd contribute then.

blizzz:
Let's continue in https://github.com/nextcloud/user_saml/pull/355

PR #313, opened 2019-03-01
Allow rewrite of UID for SSO

digitalshow:
Allow for rewrite of UIDs using preg_replace. 

An administrator can provide a pattern and a replacement parameter that get passed to PHP's preg_replace function before the UID is used to sign in to NextCloud. I use it to extract the username from a kerberos principal (pattern: `/^(.*)@REALM$/`, replacement: `$1`).


schiessle:
thanks for your contribution! :+1: Sounds like a good idea but I don't have a Kerberos setup to test it at the moment... @icewind1991 do you have a chance to review/test it?

digitalshow:
It is not limited to Kerberos, you can use it in other places where the username has to be rewritten (e.g. when the SSO mechanism passes an e-mail address or a full path from LDAP).

digitalshow:
Any news here?

skjnldsv:
@digitalshow thanks for this pr!
Could you rebase please? :)

PR #315, opened 2019-03-08
Link to portal documentation

mactrent:
The Nextcloud App Store [listing](https://apps.nextcloud.com/apps/user_saml) for this app continues to point to the URL of the last-seen `<documentation>` tag, which is a dead link.  That behavior is probably a bug in App Store, but we can overwrite the tag with an updated URL to work around it.
Edit: Sorry, this is my first Nextcloud PR, is the below enough to satisfy DCO?  I only figured I'd open the PR rather than comment to ask someone else to do it.
Signed-off-by: mactrent <3249864+mactrent@users.noreply.github.com>

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -23,6 +23,9 @@ While theoretically any other authentication provider implementing either one of
 	<types>
 		<authentication/>
 	</types>
+	<documentation>
+		<admin>https://portal.nextcloud.com/article/configuring-single-sign-on-10.html</admin>
+	</documentation>
 	<category>integration</category>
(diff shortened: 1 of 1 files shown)


schiessle:
Thanks for your PR, I'm just a bit undecided if we should link to the portal for documentation as not everybody can access it... @rullzer what do you think?

wiswedel:
Thanks for your PR, @mactrent. I've adapted it in my extended PR under https://github.com/nextcloud/user_saml/pull/356.

@schiessle I think it's fine to point to the Portal documentation, since this is clearly an enterprise feature.

PR #318, opened 2019-04-05
Support reverse proxy the NextCloud way

Toilal:
In some network configurations involving a reverse proxy, the base url generated by the underlying SAML library is not consistent with the way it's generated in NextCloud.

For example, it may generate `http://` urls instead of `https://` when the SSL Layer is handled by a proxy, even when NextCloud URLGenerator#getAbsoluteURL effectively generates `https://` urls.

This change setup SAML library to use the Server Protocol and Server Host as returned by the NextCloud Request object to build SAML urls properly.


hairmare:
Safari and Firefox both seem to currently display a security warning due to `http` only URLs in some redirects. It looks like this PR would fix the issue.

rullzer:
Right. It makes sense... let me see if I can do a test setup with this.

Toilal:
I have rebased on current master. @rullzer Have you made some tests with those changes ?

blizzz:
I am going through the open PRs and a there is a bunch of dusty ones here :sweat_smile: 

Rebased this one. I understand this is still relevant.

CarlSchwan (approved):

PR #320, opened 2019-04-09
adjust config to drone  1.0

blizzz:
* [x] master needs PHP > 7.0 https://github.com/nextcloud/docker-ci/pull/141

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,138 +1,78 @@
+kind: pipeline
+name: compatibility
+
 clone:
-  git:
-    image: plugins/git
-    depth: 1
+  depth: 1
 
-pipeline:
-  check-app-compatbility:
+steps:
+  - name: app-code-check
     image: nextcloudci/php7.1:php7.1-16
     environment:
-      - APP_NAME=user_saml
-      - CORE_BRANCH=master
-      - DB=sqlite
+      APP_NAME: user_saml
+      CORE_BRANCH: master
+      DB: sqlite
     commands:
-      # Pre-setup steps
       - wget https://raw.githubusercontent.com/nextcloud/travis_ci/master/before_install.sh
       - bash ./before_install.sh $APP_NAME $CORE_BRANCH $DB
       - cd ../server
-
-      # Code checker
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -1,138 +1,181 @@
...
-      - DB=sqlite
+      APP_NAME=user_saml
+      CORE_BRANCH=master
+      DB=sqlite
     commands:
-      # Pre-setup steps
       - wget https://raw.githubusercontent.com/nextcloud/travis_ci/master/before_install.sh
       - bash ./before_install.sh $APP_NAME $CORE_BRANCH $DB
       - cd ../server
       - php ./build/signed-off-checker.php
```
nickvergessen: @rullzer can you enable DCO instead

MorrisJobke: It is already enabled.

blizzz: ah, DCO is enabled, lemme kill that

blizzz:
the integration tests were failing already before the drone mig for some time. Let's see whether this is a low hanging fruit.

blizzz:
> /usr/drone/bin/init: line 41: php: command not found 

It comes installed with the image however. I don't get this issue locally.

> - scl enable rh-php70 bash

This should start a new bash that has the software collection or whatever it is enabled. It works locally, but apparently not with drone. When I am out of that shell, php is also unavailable. The trick seems to be to source the software collection. Temporary commit pushed – for master we need a new image with a higher PHP version.

blizzz:
> `sudo: /opt/rh/rh-php70/root/usr/bin/php: command not found `

That is fixed with the changes in  tests/integration/features/bootstrap/FeatureContext.php. At least it seems the old version is used, since there the path was hard coded. @MorrisJobke does it make sense?

But it would fail for other another reason in 7.2.

MorrisJobke:
> That is fixed with the changes in tests/integration/features/bootstrap/FeatureContext.php. At least it seems the old version is used, since there the path was hard coded. @MorrisJobke does it make sense?

Yep - makes sense 👍

blizzz:
Then, let's merge  it and see what drone say afterwards

Issue #161, opened 2017-10-02
IDP-initiated Logout not working


alferca:
Login and sp-initiated logout seems working, although I can see some errors related with mcrypt in the log:
Function mcrypt_module_open() is deprecated at \
/var\/www\/html\/nextcloud\/apps\/user_saml\/3rdparty\/vendor\/onelogin\/php-saml\/extlib\/xmlseclibs\/xmlseclibs.php#319

IDP-initiated Logout is not working.

How is affecting the mcrypt errors? is there some workaround, or is it going to be solved?

Thanks

### Steps to reproduce
1. Log in with SSO/SAML
2. Start a IDP-initiated Logout
3. Check logout is effective

### Expected behaviour
Logout initiated from idp should result in a correct logout from the SP ( nextcloud ) server.

### Actual behaviour
logout message is not processed correctly by user_saml at SP ( nextcloud ) server

### Server configuration

**Operating system**:
CentOS Linux release 7.4.1708 (Core)

**Web server:**
httpd-2.4.6-67.el7.centos.2.x86_64

**Database:**
mariadb-server-5.5.56-2.el7.x86_64

**PHP version:**
php71w-mysqlnd-7.1.9-2.w7.x86_64
mod_php71w-7.1.9-2.w7.x86_64
php71w-imap-7.1.9-2.w7.x86_64
php71w-pecl-imagick-3.4.3-1.w7.x86_64
php71w-pdo-7.1.9-2.w7.x86_64
php71w-mcrypt-7.1.9-2.w7.x86_64
php71w-cli-7.1.9-2.w7.x86_64
php71w-mbstring-7.1.9-2.w7.x86_64
php71w-pspell-7.1.9-2.w7.x86_64
php71w-ldap-7.1.9-2.w7.x86_64
php71w-pear-1.10.4-1.w7.noarch
php71w-common-7.1.9-2.w7.x86_64
php71w-xml-7.1.9-2.w7.x86_64
php71w-gd-7.1.9-2.w7.x86_64
php71w-process-7.1.9-2.w7.x86_64

**Nextcloud version:** (see Nextcloud admin page)
  - version: 12.0.3.3

**Where did you install Nextcloud from:**
https://download.nextcloud.com/server/releases/nextcloud-12.0.3.zip

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```
Enabled:
  - activity: 2.5.2
[...]

Issue #82, opened 2017-01-27
Allow HTTP-POST for single logout service


jaebischer:
My SAML IDP (Big-IP F5 APM) does only support HTTP-POST for single logout service. However Nextcloud does not accept POST requests to URL **/saml/sls**. Is it possible to support HTTP-POST requests for SAML single logout service?

LukasReschke:
That certainly makes sense and I'd be happy to review and accept a Pull Request on this. As the features that I'm working here for the SAML app is mainly influenced by [customer demand from Nextcloud GmbH customers](https://nextcloud.com/enterprise/) this is however not a priority for now.

So I won't have time to work on this on my own.

smidt:
Seems like the same issue here

fri-sch:
I have a fix (or maybe better call it a hack) for this issue: https://github.com/fri-sch/user_saml/commit/c8f6021057a558ef508c211eafc83b24833f437b

It's based on #334 and as soon as it is merged, I will send a PR for this.

PR #334, opened 2019-05-22
fix IDP-initiated Logout

DylannCordel:
cf #161 

With those websites configured to use SSO:

* `IDP`: the IDP / SSO service
* `N`: nextcloud instance
* `O` : an other website using the SSO

When user disconnect from site `O`, it send a request to `IDP`. `IDP` send a request to `N` (and all other websites where current user as a session opened) to disconnect the user on those websites too.

This PR fix this behaviour which was not working because of CRSF check which can not be valid in this scenario: validation must be done via the validation of the IDP SAML request.

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -309,6 +309,7 @@ class SAMLController extends Controller {
 	}
 
 	/**
+	 * @PublicPage
 	 * @NoAdminRequired
 	 * @NoCSRFRequired
 	 *
@@ -316,25 +317,43 @@ class SAMLController extends Controller {
 	 * @throws Error
 	 */
 	public function singleLogoutService() {
+		$isFromGS = ($this->config->getSystemValue('gs.enabled', false) &&
+					 $this->config->getSystemValue('gss.mode', '') === 'master');
+		$isFromIDP = !$isFromGS && !empty($_GET['SAMLRequest']);
 
-		$pass = $this->request->passesCSRFCheck();
-		$isGlobalScaleEnabled = $this->config->getSystemValue('gs.enabled', false);
-		$gssMode = $this->config->getSystemValue('gss.mode', '');
-		if (!$pass && $isGlobalScaleEnabled && $gssMode === 'master') {
+		if($isFromIDP) {
+			// requests comes from the IDP so let it manage the logout
+			// (or raise Error if request is invalid)
+			$pass = True ;
+		} elseif($isFromGS) {
+			// Request is from master GlobalScale
+			// Request validity is check via a JSON Web Token
 			$jwt = $this->request->getParam('jwt', '');
 			$pass = $this->isValidJwt($jwt);
+		} else {
+			// standard request : need read CRSF check
+			$pass = $this->request->passesCSRFCheck();
 		}
 
 		if($pass) {
 			$idp = $this->session->get('user_saml.Idp');
 			$auth = new Auth($this->SAMLSettings->getOneLoginSettingsArray($idp));
-			$returnTo = null;
-			$parameters = array();
-			$nameId = $this->session->get('user_saml.samlNameId');
-			$sessionIndex = $this->session->get('user_saml.samlSessionIndex');
-			$this->userSession->logout();
-			$targetUrl = $auth->logout($returnTo, $parameters, $nameId, $sessionIndex, true);
-		} else {
+			$stay = true ; // $auth will return the redirect URL but won't perform the redirect himself
+			if($isFromIDP){
+				$keepLocalSession = true ; // do not let processSLO to delete the entire session. Let userSession->logout do the job
+				$targetUrl = $auth->processSLO($keepLocalSession, null, false, null, $stay);
+			} else {
+				// If request is not from IDP, we must send him the logout request
+				$parameters = array();
+				$nameId = $this->session->get('user_saml.samlNameId');
+				$sessionIndex = $this->session->get('user_saml.samlSessionIndex');
+				$targetUrl = $auth->logout(null, [], $nameId, $sessionIndex, $stay);
+			}
+			if(!empty($targetUrl) && !$auth->getLastErrorReason()){
+				$this->userSession->logout();
+			}
+		}
+		if(empty($targetUrl)){
 			$targetUrl = $this->urlGenerator->getAbsoluteURL('/');
 		}
 


lib/Controller/SAMLController.php
```diff
@@ -316,23 +316,37 @@ public function assertionConsumerService() {
...
+			// Request is from master GlobalScale
+			// Request validity is check via a JSON Web Token
 			$jwt = $this->request->getParam('jwt', '');
 			$pass = $this->isValidJwt($jwt);
-		}
-
-		if($pass) {
+		} else {
 			$idp = $this->session->get('user_saml.Idp');
 			$auth = new Auth($this->SAMLSettings->getOneLoginSettingsArray($idp));
```
rullzer: Moving this here makes the GS mode fail right @schiessle

lib/Controller/SAMLController.php
```diff
@@ -316,23 +316,37 @@ public function assertionConsumerService() {
 	 * @throws Error
```
rullzer: I Think we need `@PublicPage` as well

Else if you log out of Nextcloud and the SLO gets called for your user and thus you get redirected back to Nextcloud you won't be logged in. right?

lib/Controller/SAMLController.php
```diff
@@ -316,23 +316,37 @@ public function assertionConsumerService() {
...
+		}
+
+		if($pass){
+			$this->userSession->logout();
+			// If request is not from IDP, we must send him the logout request
 			$parameters = array();
 			$nameId = $this->session->get('user_saml.samlNameId');
 			$sessionIndex = $this->session->get('user_saml.samlSessionIndex');
-			$this->userSession->logout();
 			$targetUrl = $auth->logout($returnTo, $parameters, $nameId, $sessionIndex, true);
```
rullzer:
```suggestion
			$targetUrl = $auth->logout(null, $parameters, $nameId, $sessionIndex, true);
```

Since your removed '$returnTo' (which was always null anyways)

rullzer:
Thanks for this @DylannCordel cool stuff

rullzer:
@schiessle @blizzz if you test with onelogin be sure to configure it so that the SLO url configured in one login points to `<server>/apps/user_saml/saml/sls` then you can see it in action

lib/Controller/SAMLController.php
DylannCordel: :open_mouth: really sorry for this newbie mistake. Thanks for your code review, it's fixed in last commits.

lib/Controller/SAMLController.php
DylannCordel: Not sure this case can append but it makes sense to avoid a login request when a non logged user want to be disconnected :). Committed.

lib/Controller/SAMLController.php
```diff
@@ -309,31 +309,46 @@ public function assertionConsumerService() {
...
+				// processSLO will do the redirection himself
+				return $auth->processSLO();
+			}
+			// If request is not from IDP, we must send him the logout request
 			$parameters = array();
 			$nameId = $this->session->get('user_saml.samlNameId');
 			$sessionIndex = $this->session->get('user_saml.samlSessionIndex');
-			$this->userSession->logout();
-			$targetUrl = $auth->logout($returnTo, $parameters, $nameId, $sessionIndex, true);
+			$targetUrl = $auth->logout(null, $parameters, $nameId, $sessionIndex, true);
```
blizzz: while at it, you can also replace `$parameters` with a simple `[]`

schiessle:
looks good from the logic, didn't tested it yet.

One security concern: We accept now logout request from everywhere without verifying if it is a valid logout request. I wonder if we could use the data from the 'SAMLRequest' parameter to verify if it is a valid request (https://github.com/nextcloud/user_saml/pull/334/files#diff-08ab8d6da860d4cc7a5530b6414b0f6eR323)?

lib/Controller/SAMLController.php
DylannCordel: done :wink:

DylannCordel:
@schiessle you right. Last commit should solve this problem. I investigate a little deeper in code and saw processSLO was able to not perform the redirect itself. Now, session logout will only be called if performSLO was a success (that means IDP request was valid).

I still have a problem with a loop login redirection when disconnecting from nextcloud due to `if(!$this->request->passesStrictCookieCheck()) {`, from `beforeController` in `SecurityMiddleWare.php` but I believe it's not related to this issue. I'll confirm that today and create an other issue and PR if needed.

skjnldsv:
Hey there!
What is the status?
Can I help somehow? :)

DylannCordel:
just merged with master to stay up to date for testing.

skjnldsv:
@DylannCordel you need to sign your commits 😉

skjnldsv:
> just merged with master to stay up to date for testing.

It would have been better to rebase to master instead of pulling master :)

DylannCordel:
just corrected noise I introduced in commits :wink:

skjnldsv:
Testing on forks seems broken, I created a potential fix https://github.com/nextcloud/user_saml/pull/338

skjnldsv:
Tests are greeeen :D

skjnldsv:
@DylannCordel do you think you could add a few tests for your changes?

fri-sch:
Thanks a lot @DylannCordel for tackling this issue. I tested in on my instance and it works just fine!  
I had one more issue with my IdP (Keycloak), that was already reported here: #82. I prepared a fix/hack for this on top of this change (https://github.com/fri-sch/user_saml/commit/c8f6021057a558ef508c211eafc83b24833f437b) and will send a MR for it as soon as this one was merged.

skjnldsv:
@DylannCordel let's get this in?

DylannCordel:
@fri-sch thanks for feed-back.

@skjnldsv yes I can check how tests works in Nextcloud and try to add some soon. Concerning modifications of @fri-sch it looks good. How do you want to proceed ? Do I cherry-pick its commit to add it on the PR (or maybe @fri-sch can PR into webu/user_saml repo ?) or do you want two distinct PR on nextcloud/user_saml ?

skjnldsv:
@DylannCordel can you open a pr against this one?

DylannCordel:
@skjnldsv I'm sorry, I don't understand what I must do:

1. is it for tests you would like a new PR based on this one ? 
2. or is it for @fri-sch 's commit ? In this case, is'n't it better if it's @fri-sch who creates a PR based on my branch ?

Sorry for time loss :flushed:

skjnldsv:
> Sorry for time loss

No worries :)
I think I misunderstood. I thought you were asking if you should add the tests here or in another pr (which is fine to have here as well, but if you really wanted to separate the two, you could have created another pr to merge in thi sone).


Regarding @fri-sch's pr, both are fine really.
We can merge this one and have another pr for the rest of the work :)

skjnldsv:
Done, let's not wait too long :)
@fri-sch could you open your pull request?

fri-sch:
@skjnldsv I opened the PR: #340

PR #335, opened 2019-05-24
Load a timezone file if no timezone is set

rullzer:


rullzer:
Yay green https://drone.nextcloud.com/nextcloud/user_saml/1232

PR #338, opened 2019-06-07
Fix testing on forks

skjnldsv:
Branch won't exist if this is coming from a fork

This should fix it


skjnldsv:
Hum, lots of       `Notice: Undefined offset: -1 in features/bootstrap/FeatureContext.php line 128`

PR #340, opened 2019-06-17
Handle SLO logout requests from IdP via POST

### LINKED ISSUE #82: Allow HTTP-POST for single logout service

fri-sch:
Some IdPs send their SLO logout requests via POST. To handle
them we need to add an entry in the routing table.
Further, we need to hack around the issue, that php-saml only
handles GET by copying the request from $_POST to $_GET.

fix #82

Signed-off-by: Frieder Schrempf <frieder.schrempf@online.de>

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -319,6 +319,13 @@ class SAMLController extends Controller {
 	public function singleLogoutService() {
 		$isFromGS = ($this->config->getSystemValue('gs.enabled', false) &&
 					 $this->config->getSystemValue('gss.mode', '') === 'master');
+
+		// Some IDPs send the SLO request via POST, but OneLogin php-saml only handles GET.
+		// To hack around this issue we copy the request from _POST to _GET.
+		if(!empty($_POST['SAMLRequest'])) {
+			$_GET['SAMLRequest'] = $_POST['SAMLRequest'];
+		}
+
 		$isFromIDP = !$isFromGS && !empty($_GET['SAMLRequest']);
 
 		if($isFromIDP) {
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -49,6 +49,12 @@ return [
 			'verb' => 'GET',
 		],
 		[
+			'name' => 'SAML#singleLogoutService',
+			'url' => '/saml/sls',
+			'verb' => 'POST',
+			'postfix' => 'slspost',
+		],
+		[
 			'name' => 'SAML#notProvisioned',
 			'url' => '/saml/notProvisioned',
 			'verb' => 'GET',
--- a/tests/unit/AppInfo/RoutesTest.php
+++ b/tests/unit/AppInfo/RoutesTest.php
@@ -55,6 +55,12 @@ class Test extends TestCase  {
 					'verb' => 'GET',
 				],
 				[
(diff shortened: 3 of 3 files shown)


lib/Controller/SAMLController.php
```diff
@@ -319,6 +319,13 @@ public function assertionConsumerService() {
 	public function singleLogoutService() {
 		$isFromGS = ($this->config->getSystemValue('gs.enabled', false) &&
 					 $this->config->getSystemValue('gss.mode', '') === 'master');
+
+		// Some IDPs send the SLO request via POST, but OneLogin php-saml only handles GET.
+		// To hack around this issue we copy the request from _POST to _GET.
+		if(!empty($_POST['SAMLRequest'])) {
```
ChristophWurst: use `$this->request->getParam(...)` instead. it's an abstraction of the `$_GET` and `$_POST` vars that allows unit testing

fri-sch: The problem is that I need to check for $_POST here explicitly to copy the value to $_GET for the underlying php-saml lib. If I would check via `getParam()`, I wouldn't know whether the data was passed via GET or POST and I wouldn't know if I need to fill $_GET.

The only thing I can imagine would be to set $_GET unconditionally like:
```
$_GET['SAMLRequest'] = $this->request->getParam('SAMLRequest');
```

Is that what you mean?

ChristophWurst: I'm sorry, I didn't know about the internal handling of superglobals in php-saml. In that case I guess it's fine to access the globals directly.
It is possible to access only get or post vars via IRequest::get($name) and IRequest::post($name) but it will throw an exception if you access get on post requests and vice versa.

skjnldsv: @ChristophWurst so leave it like that?

ChristophWurst: yes!

skjnldsv: Review? :D

ChristophWurst: Yes, someone who knows the SAML code should have a look :)

ChristophWurst: https://github.com/nextcloud/user_saml/pull/340#event-2423557266 :see_no_evil:

Ornias1993:
Did anyone actually test this?
The whole " SLO logout requests from IdP via POST" doesn't seem to have been working, both before and after this patch.

Nextcloud at the very least doesn't close the session for the logged-in user.

fri-sch:
@Ornias1993 When I tested it back then I though it worked, but maybe I missed something. I didn't dive into this very deeply and I currently cannot remember the details. Thanks for following up on this though.

Ornias1993:
@fri-sch Thanks for the reply! :)
I was refering to #455 mostly.

With the "new" way nextcloud handles auth (introduced a few years back or so, based on oauth), there is no way to do idp initiated auth.

It's kinda insane when you consider they actually ask money for access to the docs, for a feature that isn't actually working. I think thats pretty close to fraud to be honest.

PR #346, opened 2019-07-17
Fix reloading NC after logging out from IDP

LEDfan:
Related to #345 
Using: Nextcloud master (17) user_saml master and keycloak `4.8.3.Final`.

## Steps to reproduce:

 1. Login on Nextcloud using Keycloak
 2. Go to the account settings of Keycloak e.g. `https://idp/auth/realms/my-realm/account/`
 3. Click the logout button
 4. Reload an existing tab

## Expected

After the #334 and #340 it is expected to be logged out and redirected to the login page of the SSO again.

## Actual Result

You get a redirect loop:

Selection_0052

The third request doesn't have a valid CRSF token and fails.

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -160,6 +160,7 @@ class SAMLController extends Controller {
 	/**
 	 * @PublicPage
 	 * @UseSession
+	 * @NoCSRFRequired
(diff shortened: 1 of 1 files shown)


blizzz (changes requested):
This annotation is already present two lines below

PR #354, opened 2019-09-19
Handle ObjectGUID claims

blizzz:
They are base64-encoded binary string. 

Also drops Nc14 support, due to failing integration tests with unclear cause. It is EOL anyhow. Version bump to 2.5 included.


blizzz:
meanwhile i think the failing tests for 14 were hiccups.. Anyway, let the dead rest in peace, shall we?

Does it look OK for you?

blizzz:
Got positive feedback from a customer  👍

lib/UserBackend.php
```diff
@@ -691,4 +691,46 @@ public function updateAttributes($uid,
...
+		if(preg_match('/^[a-f0-9]{8}(-[a-f0-9]{4}){4}[a-f0-9]{8}$/i', $candidate) === 1) {
+			$uid = $candidate;
+		}
+		return $uid;
+	}
+
+	/**
+	 * @see \OCA\User_LDAP\Access::convertObjectGUID2Str
+	 */
+	public function convertObjectGUID2Str($oguid) {
```
skjnldsv: Does it needs to be public?

PR #355, opened 2019-10-08
Add dedicated group backend

blizzz:
continuation of #311 

* [x] migration strategy
* [ ] gid collision handling
* [ ] automatically delete groups without members 
* [ ] ensure admins cannot (un)assign members, create or delete saml groups
* [ ] tests


blizzz:
I am still considering a good migration strategy.

Type | Manual / Auto | code lifetime | description
-- | -- | -- | --
Aggressive | Auto | temporary | any group that comes from SAML will be migrated from local to saml backend, except "admin". Local groups created after the saml group backend was introduced should not be touched. Business logic can be dropped in when 17 or 18 reaches EOL.
From Snapshot | Auto | temporary | current local groups are remembered (but admin). saml groups coming in that are being part will be migrated. after e.g. 60 days, forget remaining groups. Business logic can be dropped in when 17 or 18 reaches EOL.
Notice | Manual | both | when an incoming saml group exists in local backend, notify admin. have an occ command. Local groups created after the saml group backend was introduced should not be touched. Business logic can be dropped in when 17 or 18 reaches EOL, apart from the migration command.
Opt-in | Manual | permanent | provide an occ command to migrate local groups to saml backend. Dropping functionality might look weird (as above)

My preference is to go with option 2. Like in 1. a list needs to be kept, only we track the candidates to migrate, i.e. whitelist, not blacklist.  It has the charm to not burden the admin with more handwork. By doing it automatically, it can cause a false positive to be migrated. However it would be trivial to bring it back (move it to the other DB table again), an SQL query or occ command can be provided if really necessary. 

When moving groups, the gid must stay stable, therefore adding a prefix or suffix afterwards is not an option – there is no renaming of group ids. This can be done for newly created groups, but would look inconsistent. So i'd stay away of it, too.

What to do on name collisions?

1. When a local groups is being created, an error will appear and the admin has to pick another name. That's fine, no action needed.
2. When a SAML group appears and collides with an existing name, the options are
  * use the local group (NO - that's our problem :) )
  * add a prefix or suffix, string or number
  * issue an error to the admin

Changing the group is inevitable in that case, but to recognize those groups we have to map the divergent gid to the original saml group id. That well be another column to add to the DB table.

Sending error messages alone is not helpful in any way, causes a delay, and hazzle. Software should make this simpler instead.

Deltachaos:
@blizzz i would also prefer option 2. If this is agreed i could implement this. I could also add the emtpy group removal. The only thing i am not sure about are the tests. I have no idea how i could get them running localy as the plugin is only testable within a nextcloud installation.

Future i have seen that you have removed the option of having a group prefix. I would like to keep this because it is a handy feature for an admin to prevent group conflicts if he has added are multiple SAML providers.

Kosyne:
It's been a while since there was activity here-- is this still in the works or being considered? If it means anything, I know that myself and a few others who area really looking forward to this.

blizzz:
> It's been a while since there was activity here-- is this still in the works or being considered? 

Yes, but a bit postponed.

> If it means anything, I know that myself and a few others who area really looking forward to this.

I have no doubts about this, just honestly, expressing it does not make it implemented faster.

JonathanTreffler:
> Yes, but a bit postponed.

Any updates ?

JonathanTreffler:
closed in favor of #536 (continuation)

JonathanTreffler:
> I have no idea how I could get them running localy as the plugin is only testable within a nextcloud installation.

The question was quite a while ago, but still :)
https://docs.nextcloud.com/server/latest/developer_manual/getting_started/devenv.html
(I wrote a [Gitpod config](https://github.com/JonathanTreffler/nextcloud-app-gitpod), that automatically deploy such a dev environment with vs code in the cloud)

PR #356, opened 2019-10-10
added link to portal documentation

### LINKED ISSUE #244: Documentation is missing
Hey All,

Thanks for providing this great feature.
As I am relativley new to SSO and SAML, I would love to reach back to some documentation.
Anything I find online is either outdated or "file not found" for my version of Nextcloud 13.

Where would you recommend me to go looking?

I am trying to setup SSO with SAML against an Azure Active Directory (WestEurope) with a Nextcloud 13 installed from tar-package on an Ubuntu with NFS (Azure File Storage) and a mysql-db on localhost.

Looking forward to your responses.

Best,
Enno

wiswedel:
fixes #244 
continues https://github.com/nextcloud/user_saml/pull/315

Signed-off-by: Sascha Wiswedel <sascha.wiswedel@nextcloud.com>

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -23,6 +23,9 @@ While theoretically any other authentication provider implementing either one of
 	<types>
 		<authentication/>
 	</types>
+	<documentation>
+		<admin>https://portal.nextcloud.com/article/configuring-single-sign-on-10.html</admin>
+	</documentation>
 	<category>integration</category>
 	<website>https://github.com/nextcloud/user_saml</website>
 	<bugs>https://github.com/nextcloud/user_saml/issues</bugs>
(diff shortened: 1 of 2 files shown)


wiswedel:
anybody fancy reviewing?

appinfo/info.xml
```diff
@@ -23,6 +23,9 @@ While theoretically any other authentication provider implementing either one of
 	<types>
 		<authentication/>
 	</types>
+	<documentation>
+		<admin>https://portal.nextcloud.com/article/configuring-single-sign-on-10.html</admin>
```
skjnldsv: Is there no unique direct link?
I think we should not use text like that that can potentially change? :thinking:

wiswedel: Not that I'm aware of. What do you suggest instead?

skjnldsv: Nothing if there is no alternative :p

PR #366, opened 2019-11-28
Add counting to the user backend

rullzer:
This will allow reporting to also list the number of SAML users on the
instance.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -6,7 +6,7 @@ clone:
 
 steps:
   - name: app-code-check
-    image: nextcloudci/php7.1:php7.1-16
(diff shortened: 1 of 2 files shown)


rullzer:
CI is not fully happy but that seems unrelated. Lets have a look separately.

blizzz:
tests bricked by server changes:

```
PHP Fatal error: Declaration of OCA\User_SAML\Tests\AppInfo\ApplicationTest::setUp() must be compatible with Test\TestCase::setUp(): void in /drone/server/apps/user_saml/tests/unit/AppInfo/ApplicationTest.php on line 31
```

rullzer:
Ah right we need to fix those...

PR #368, opened 2019-12-05
Update master php testing versions

nickvergessen:


blizzz:
thanks

blizzz:
Ah, we have to abandon older releases here as well, due to the phpunit switch to 8.

PR #389, opened 2020-01-29
bump version for 2.4.1

blizzz:
and adjust CI


blizzz:
now integration tests became not so reliable. first run, all went through. Sometimes we see `ERROR [net.shibboleth.idp.attribute.resolver.dc.ldap.impl.ConnectionFactoryValidator:95] - Connection factory validation failed `, but within the last three runs, each was finishing successfully twice…

PR #391, opened 2020-01-29
[stable-2.4] Handle ObjectGUID claims

blizzz:
backport of #354


blizzz:
ldap wasn't ready in time, but other int tests went through → fine

PR #396, opened 2020-02-05
Add setting to specify a different signature algorithm

### LINKED ISSUE #310: Allow selecting of signature algorithm
Some older IdPs do not support the default sha256 used in the lib to sign the requests with.
We should allow selecting the signature algorith.

juliushaertl:
Add a separate setting for selecting a different signature algorithm.

@rullzer Sorry to disappoint you a bit with no vue rewrite yet, but I've put it on my list of things for some fun hacking :wink: 

Fixes #310 

The list of possible values is defined in [php-saml](https://github.com/onelogin/php-saml):

        http://www.w3.org/2000/09/xmldsig#rsa-sha1
        http://www.w3.org/2000/09/xmldsig#dsa-sha1
        http://www.w3.org/2001/04/xmldsig-more#rsa-sha256
        http://www.w3.org/2001/04/xmldsig-more#rsa-sha384
        http://www.w3.org/2001/04/xmldsig-more#rsa-sha512

### DIFF
--- a/lib/SAMLSettings.php
+++ b/lib/SAMLSettings.php
@@ -119,6 +119,7 @@ class SAMLSettings {
 				'wantXMLValidation' => ($this->config->getAppValue('user_saml', $prefix . 'security-wantXMLValidation', '0') === '1') ? true : false,
 				'requestedAuthnContext' => false,
 				'lowercaseUrlencoding' => ($this->config->getAppValue('user_saml', $prefix . 'security-lowercaseUrlencoding', '0') === '1') ? true : false,
+				'signatureAlgorithm' => $this->config->getAppValue('user_saml', $prefix . 'security-signatureAlgorithm', null)
 			],
 			'sp' => [
 				'entityId' => $this->urlGenerator->linkToRouteAbsolute('user_saml.SAML.getMetadata'),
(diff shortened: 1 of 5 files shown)


lib/SAMLSettings.php
```diff
@@ -119,6 +119,7 @@ public function getOneLoginSettingsArray($idp) {
 				'wantXMLValidation' => ($this->config->getAppValue('user_saml', $prefix . 'security-wantXMLValidation', '0') === '1') ? true : false,
 				'requestedAuthnContext' => false,
 				'lowercaseUrlencoding' => ($this->config->getAppValue('user_saml', $prefix . 'security-lowercaseUrlencoding', '0') === '1') ? true : false,
+				'signatureAlgorithm' => $this->config->getAppValue('user_saml', $prefix . 'security-signatureAlgorithm', null)
```
rullzer: Doe null fall back to the current default?

juliushaertl: Yes, that way we use the default from php-saml https://github.com/nextcloud/user_saml/blob/master/3rdparty/vendor/onelogin/php-saml/src/Saml2/Settings.php#L395-L397

PR #414, opened 2020-03-28
Better redirect user to homepage instead of showing wired error

Deltachaos:
Empty return displays only the word "null" in the browser. Users who
land here in case of an error are completely lost. We should better
redirect to the homepage in this case.

Signed-off-by: Maximilian Ruta <mr@xtain.net>


svenseeberg:
Not sure why the build is marked as "failed". Is the result already deleted? Having this merged would be cool.

blizzz:
@Deltachaos thanks for the PR. I opened https://github.com/nextcloud/user_saml/pull/446 as a replacement, which goes a step further and adds debug logging.

PR #430, opened 2020-06-19
Make work with posts and cookies again

### LINKED ISSUE #433: SAML stops working after upgrade to NC 19.
SAML is no longer working after upgrade to NC 19. Empty page with `Null` is displayed after login succeeds through Google G. Suite.

Commenting out the following code solves the problem as long as multiple IDPs are not configured in NextCloud (when more than 1 IDP is configured - the plugin doesn't know which one to use as $idp is empty):
```
//if(is_null($AuthNRequestID) || $AuthNRequestID === '' || is_null($idp)) {
//	return;
//}
```

I was digging through the code and discovered that values in the Session are not preserved - both `user_saml.AuthNRequestID` and `user_saml.Idp` are empty when `assertionConsumerService` is called. They were initialized properly during `login($idp)` call.

* Why should `AuthNRequestID` be stored in the Session in the first place, as SAML protocol POST to ACP call is providing this value. Using the value from SAML POST call would also solve [this](https://github.com/nextcloud/user_saml/issues/208) problem altogether.
* The problem also referenced [here](https://help.nextcloud.com/t/idp-iniatied-saml-logging-null-webpage-text-while-sp-initiated-works-fine/68116/9) 

### Server configuration

**Operating system**: GCP Container OS with docker-compose based setup
**Web server:** Nginx reverse proxy
**Database:** MySQL
**PHP version:**
**Nextcloud version:** 19.0.0
**Updated from an older Nextcloud/ownCloud or fresh install:** Updated from 18.0.5

rullzer:
Requires https://github.com/nextcloud/server/pull/21479 to fully work.
Basically don't save this info in the session (which is lax by default
starting with NC19 but also soon with new chromes and firefox). We now
save it is a cookie that is set to None. This is the best we can do I
think.

fixes #433

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -37,6 +37,7 @@ use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserManager;
 use OCP\IUserSession;
+use OCP\Security\ICrypto;
 use OneLogin\Saml2\Auth;
 use OneLogin\Saml2\Error;
 use OneLogin\Saml2\Settings;
@@ -61,6 +62,10 @@ class SAMLController extends Controller {
 	private $logger;
 	/** @var IL10N */
 	private $l;
+	/**
+	 * @var ICrypto
+	 */
+	private $crypto;
 
 	/**
 	 * @param string $appName
@@ -85,7 +90,8 @@ class SAMLController extends Controller {
 								IURLGenerator $urlGenerator,
 								IUserManager $userManager,
 								ILogger $logger,
-								IL10N $l) {
+								IL10N $l,
+								ICrypto $crypto) {
 		parent::__construct($appName, $request);
 		$this->session = $session;
 		$this->userSession = $userSession;
@@ -96,6 +102,7 @@ class SAMLController extends Controller {
 		$this->userManager = $userManager;
 		$this->logger = $logger;
 		$this->l = $l;
+		$this->crypto = $crypto;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


lib/Controller/SAMLController.php
```diff
@@ -244,8 +264,22 @@ public function getMetadata($idp) {
...
+		$cookie = $this->request->getCookie('saml_data');
+		if ($cookie === null) {
+			return;
+		}
+
+		// Base64 decode
+		$cookie = base64_decode($cookie);
+
+		// Decrypt and deserialize
+		$cookie = $this->crypto->decrypt($cookie);
```
kesselb, line 277: Mind to update the throws block? `decrypt` might throw a `\Exception`. I guess that could happen if one sends a self crafted saml_data cookie ;)

rullzer: Good catch yes. We should just catch it I guess and then return an error

lib/Controller/SAMLController.php
```diff
@@ -175,9 +181,22 @@ public function login($idp) {
...
+					'Idp' => $idp
+				]);
+
+				// Encrypt it
+				$data = $this->crypto->encrypt($data);
+
+				// And base64 encode it
+				$data = base64_encode($data);
+
+				$response->addCookie('saml_data', $data, null, 'None');
```
blizzz, line 199: Note [to myself], before merging, we ought to branch off (and better switch to regular stable branches, as we have them in most other apps).

rullzer: I don't really see why.
The code here should work also work on older releases. else we end up with the backport mess for every release.

blizzz: Okay. I expected a Warning when the parameter count mismatches.

rullzer: ah mmm right...

blizzz: It does not complain when there are too many, only when there are too few. So it works. If awkward.

lib/Controller/SAMLController.php
MorrisJobke: 🏓

blizzz:
It was merged three years ago. There is nothing to do. 

For discussions I would like to ask you to raise your question in the forums: https://help.nextcloud.com

PR #436, opened 2020-07-24
Ignore the oauth token path in login with the environment variables

rullzer:


blizzz (approved):
:elephant:

PR #458, opened 2020-09-25
Handle failing SLO

rullzer:
If the SLO throws an error we should catch it. This is so that we do not
show an error page. We should also still logout the current session.


juliushaertl (approved):
Makes sense :+1:

PR #463, opened 2020-10-05
Fix login flow with SAML

rullzer:
Because of the strict samesite cookies SAML fails with the login flow.
Because the post that comes back is not transfering the proper cookies
to use the same session. Hence the token in use gets lost etc.

Now we store this all (encrypted) in a cookie. So that when we come back
we can restore the proper session.

FAQ:

* Is it elegant?
  Nope!
* Does it work?
  Yes!

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -23,6 +23,8 @@
 namespace OCA\User_SAML\Controller;
 
 use Firebase\JWT\JWT;
+use OC\Core\Controller\ClientFlowLoginController;
+use OC\Core\Controller\ClientFlowLoginV2Controller;
 use OCA\User_SAML\Exceptions\NoUserFoundException;
 use OCA\User_SAML\SAMLSettings;
 use OCA\User_SAML\UserBackend;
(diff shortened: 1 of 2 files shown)


schiessle:
Just to be on the save side: SSO with environment variables like OAuth or Kerberos are not affected?

rullzer:
> Just to be on the save side: SSO with environment variables like OAuth or Kerberos are not affected?

no.
There we don't set cookies like this

lib/Controller/SAMLController.php
```diff
@@ -197,7 +212,7 @@ public function login($idp) {
 				// And base64 encode it
 				$data = base64_encode($data);
 
-				$response->addCookie('saml_data', $data, null, 'None');
+				$response->addCookie('saml_data', $data, 1800, 'None');
```
juliushaertl, line 215:
```suggestion
				$expiry = new \DateTime();
				$expiry->add(new \DateInterval('PT30M'));
				$response->addCookie('saml_data', $data, $expiry, 'None');
```

rullzer: ha ah shit you are right

juliushaertl (approved):

lib/Controller/SAMLController.php
rullzer: Let me set it to null again actually to prevent issues.

PR #476, opened 2020-11-02
redirects to homepage instead showing error on blank page

blizzz:
one way to easily reproduce it:

* log into Nextcloud via SAML
* in the same browser, open again https://my.nxtcld.srv/index.php/apps/user_saml/saml/selectUserBackEnd?redirectUrl=

Before the patch, you'll see a blank page with  the error.

With it, you are being redirected to the homepage of your Nextcloud instance.


juliushaertl (approved):

juliushaertl:
backports?

PR #480, opened 2020-11-04
fix missing user_saml.Idp session value which SAMLSettings rely on

blizzz:
quick fix candidate for a regression introduced in https://github.com/nextcloud/user_saml/pull/430

To figure out the correct settings to fetch during session lifetime, `SAMLSettings::getPrefix` relies on the presence of the session stored value that got lost here. 

* [x] draft, because i have not tested it much yet

### DIFF
--- a/js/admin.js
+++ b/js/admin.js
@@ -351,6 +351,7 @@ $(function() {
 			// Checks on each request whether the settings make sense or not
 			$.ajax({
 				url: OC.generateUrl('/apps/user_saml/saml/metadata'),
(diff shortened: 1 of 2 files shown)


blizzz:
Also fixes the wrong meta data being loaded on the settings ui due to missing idp parameter.

blizzz:
And for the record, it is technically not a regression, but rather a left over of the original issue.

juliushaertl (approved):

PR #486, opened 2020-11-23
fixes provisioning of userids from  encoded (objectguid) values

blizzz:
- is more tolerate when decoding, uuid structure is still tested later
- ensures the uid is resolved on getCurrentId()


juliushaertl (approved):
Makes sense :+1:

PR #490, opened 2020-11-27
Remove obsolete update.php

ChristophWurst:
They call me the 007 of SAML

0 … knowledge about what this code does
0 … manual tests
7… possible ways this could break

On a more serious note I'm not sure if this is even necessary anymore. But we can't quite assume that everyone updated to a newer version and this had been run before. So better have this in any case.


lib/Migration/SetDefaultType.php
```diff
@@ -0,0 +1,55 @@
...
+		return 'Set default for SAML desktop authentication when updating from v1.2.0 or earlier';
+	}
+
+	public function run(IOutput $output): void {
+		$installedVersion = $this->config->getAppValue('user_saml', 'installed_version');
+
+		// Versions below 1.2.1 use SAML by default for the desktop client, this default
+		// has been changed with 1.2.1. To not break existing installations the value gets
+		// manually changed on update.
+		if (version_compare($installedVersion, '1.2.1', '<')) {
```
nickvergessen, line 51:
```suggestion
		if (version_compare($installedVersion, '1.2.2', '<')) {
```

lib/Migration/SetDefaultType.php
```diff
@@ -0,0 +1,55 @@
...
+			$this->config->setAppValue('user_saml', 'general-use_saml_auth_for_desktop', '1');
```
nickvergessen, line 52:
```suggestion
			$this->config->setAppValue('user_saml', 'type', 'saml');
```

blizzz:
We can simply remove it without replacement:

Screenshot_20201127_145414

MorrisJobke (approved):

PR #497, opened 2020-12-23
support for encryption

immerda:
This allows the idp to return a stable user_secret, which can be
used by e.g. the encryption app in lieu of the password.

We would very, very much like to have this feature in production!

Requires https://github.com/nextcloud/server/pull/24831 to work.

The problem we found is that this app uses `handleApacheAuth()` to initialize the session,
which overrides the encryption app keystore, by emiting
a [login event](https://github.com/nextcloud/server/blob/3a30ac495bd9663ef8233aacc8ce5b21a8adce67/lib/private/legacy/OC_User.php#L180).

We think the easiest approach is to [prevent the key from being overwritten](https://github.com/nextcloud/server/pull/24831).

Let me know if you don't like the approach. An alternative would be to modify `handleApacheAuth` to query the backend for a user secret...

### DIFF
--- a/lib/Settings/Admin.php
+++ b/lib/Settings/Admin.php
@@ -130,7 +130,11 @@ class Admin implements ISettings {
 				'type' => 'line',
 				'required' => true,
 			],
-
+			'user_secret_mapping' => [
+				'text' => $this->l10n->t('Attribute to use as user secret e.g. for the encryption app.'),
+				'type' => 'line',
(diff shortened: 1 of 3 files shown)


immerda:
Unfortunately this does not seem to work when using the android app or when authenticating through app passwords.

I am fairly stuck trying to figure out where the token for the app is created. Any pointers?

PR #500, opened 2021-01-12
Allow customer directlogin text

rullzer:
Some people seem to want to have a custom direct login text. This allows
them to set it. For now only via occ. But maybe some day we also add a
GUI component to it.


blizzz (commented):
does not help with multiple saml configurations

blizzz:
> does not help with multiple saml configurations

Oh wait, I just realized this is renaming the direct login, and not the SAML providers themselves

rullzer:
> > does not help with multiple saml configurations
> 
> Oh wait, I just realized this is renaming the direct login, and not the SAML providers themselves

exactly :)

blizzz (approved):

juliushaertl (approved):
Yay :tada:

PR #501, opened 2021-01-12
Bump doctrine types for 21 support

rullzer:


ChristophWurst (changes requested):
This is brilliant

But I like [this](https://github.com/nextcloud/server/pull/25089)

PR #505, opened 2021-01-20
make testEncodedObjectGUID more robust against false positives

### LINKED ISSUE #504: UID mapping not working properly
### Steps to reproduce
1. Enable SAML SSO and attribute UID mapping, use "email" as UID
2. Login

### Expected behaviour

I am logged in with the UID set to my user name.

### Actual behaviour

A new user with the UID "796B776A-2BAB-B79D-289E-CBA5B629E06A" is created instead. I'm not yet sure what it depends on, because I did not observe this behavior on another instance with a very similar configuration. Further, it seems to have surfaced suddenly. A debug log correctly lists the attributes passed to user_saml, I just can't share it due to privacy reasons.

### Server configuration

NC 20.0.5. with user_saml 3.3.3, auto-provisioning of SAML users allowed.

PHP: 7.3.24

### Possible cause

Commit https://github.com/nextcloud/user_saml/commit/9f53230eb6e1c97b006be30fb83fc230b09534eb changed the way `testEncodedObjectGUID` works, the parameter `strict` to `base64_decode` is set to false. I'm not sure how to interpret the description in the PHP manual, but IMHO `base64_decode` will almost never return false if `strict` is also false, because it simply skips characters it can't decode. 
If I change strict from false to true, I can log in: https://github.com/nextcloud/user_saml/blob/b406b6bdfa7515690695be400348161f6fbfd8b5/lib/UserBackend.php#L705

blizzz:
There are some collisions possible. The received string is now tested against allowed characters as put down in https://tools.ietf.org/html/rfc4648#section-4

* [ ] https://github.com/nextcloud/docker-ci/pull/261
* fixes #504

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -71,7 +71,7 @@ clone:
 
 steps:
   - name: integration-tests-master
-    image: nextcloudci/user_saml_shibboleth-php7.3:user_saml_shibboleth_php7.3-1
+    image: nextcloudci/user_saml_shibboleth-php7.3:user_saml_shibboleth_php7.3-2
     environment:
       CORE_BRANCH: master
     commands:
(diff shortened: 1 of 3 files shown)


rullzer (approved):

rullzer:
Backport as well I guess

blizzz:
> Backport as well I guess

o'course

blizzz:
bumped the integration test in the search of green lights. The latter commit ought not to be backported thus.

PR #417, opened 2020-04-09
sanitize and test user id received from IdP, if original does not match

blizzz:
… to cope with adventurous LDAP configs

It does not fix that autoprovisioning would pass on otherwise not allowed characters in its user id.

PR #508, opened 2021-01-26
sanitize and test user id received from IdP, if original does not match

blizzz:
Continuation of #417


rullzer (approved):
Didn't test but looks sane

PR #513, opened 2021-02-15
Add logging for SLO errors

LukasReschke:
To make debugging SLO errors easier, this adds logging for any
encountered error in that phase.

This is similar to the logging already done on the ACS handling.

Signed-off-by: Lukas Reschke <lukas@statuscode.ch>


LukasReschke:
@rullzer, @juliushaertl WDYT? That should make debugging SLO issues significantly easier. Right now in the case of SLO failures, you just end up in the Nextcloud Files app and there's no information what went wrong anywhere.

rullzer (approved):
Yeah makes sense to log this high.

PR #514, opened 2021-02-16
[stable-3.2] update composer files

blizzz:
As would be the case with make appstore (apart of the `"dev": "true"`) line.


LukasReschke (approved):

blizzz:
or it is my too new composer here https://github.com/nextcloud/docker-ci/blob/master/php7.2/Dockerfile#L27-L29

blizzz:
> or it is my too new composer here https://github.com/nextcloud/docker-ci/blob/master/php7.2/Dockerfile#L27-L29
> 
> ```
> PHP Warning:  require(/drone/server/apps/user_saml/3rdparty/vendor/composer/platform_check.php): failed to open stream: No such file or directory in /drone/server/apps/user_saml/3rdparty/vendor/composer/autoload_real.php on line 25
> ```

yes

PR #515, opened 2021-02-16
check that all depenencies are installed

blizzz:
otherwise we see it only on packaging


rullzer (approved):

blizzz:
gonna do manual backports (for version changes) later on.

Issue #187, opened 2018-02-27
Extra SAML attributes


soumplis:
It would be nice if we could populate / update a user's profile info by releasing extra attributes from the IdP  to nextcloud.

imacks:
+1

Populating the avatar would be awesome

Lumrenion:
I would like to populate the avatar url from SAML as well. It might be helpful to take a look at user_dap and see how they implemented this. Mind, if I dig into this?

PR #520, opened 2021-03-23
Populate avatar from SAML attribute

Lumrenion:
This PR addresses #187. 
- In admin configuration, a new configuration field avatar_mapping is added.
- If avatar_mapping configuration field is not empty, the user wont be able to upload an image in the nextcloud profile settings
- The avatar_mapping field holds a publicly available URL to the profile image of the user. As usual, the image will be downloaded to nextcloud when the user logs in via user_saml.


Lumrenion:
It is my first code contribution to a Nextcloud project, so I am not sure, how the merging process works. Do I need to do something for this PR being merged?

lib/UserBackend.php
```diff
@@ -21,6 +21,8 @@
 
 namespace OCA\User_SAML;
 
+use OC\User\Backend;
+use OCA\User_LDAP\FilesystemHelper;
```
blizzz, line 25: Don't use this – it belongs to another app and might not be available, thus breaking user_saml.

l10n/de.json
```diff
@@ -29,6 +29,7 @@
     "Attribute to map the quota to." : "Attribut, dem das Kontingent zugeordnet werden soll.",
     "Attribute to map the users groups to." : "Attribut, dem die Gruppen des Benutzers zugeordnet werden sollen.",
     "Attribute to map the users home to." : "Attribut dem das zu Hause des Benutzers zugeordnet werden soll.",
+    "Attribute to map the users avatar to." : "Attribut, dem das das Profilbild des Benutzers zugeordnet werden soll.",
```
blizzz, line 32: Please don't change any translations here directly – they are being done on transifex, changes would be overwritten.

lib/UserBackend.php
```diff
@@ -81,6 +83,28 @@ public function __construct(
...
+    public function canChangeAvatar($uid) {
+        if (!$this->implementsActions(Backend::PROVIDE_AVATAR)) {
+            return true;
+        }
+        try {
+            if (empty(trim($this->getAttributeKeys('saml-attribute-mapping-avatar_mapping')[0]))) {
+                return true;
+            }
+        } catch(\InvalidArgumentException $e) {
+            return true;
```
blizzz, line 102: :thinking: false?

Also, are you sure you don't want to check the return of `getAttributeValue` instead a few lines above? You can save the if and return the result of the comparison.

lib/UserBackend.php
```diff
@@ -181,6 +205,7 @@ public function implementsActions($actions) {
 		$availableActions |= \OC\User\Backend::GET_DISPLAYNAME;
 		$availableActions |= \OC\User\Backend::GET_HOME;
 		$availableActions |= \OC\User\Backend::COUNT_USERS;
+        $availableActions |= \OC\User\Backend::PROVIDE_AVATAR;
```
blizzz, line 208: indentation (cf. https://docs.nextcloud.com/server/latest/developer_manual/getting_started/codingguidelines.html)

lib/UserBackend.php
```diff
@@ -636,6 +661,14 @@ public function updateAttributes($uid,
 			$newGroups = null;
 		}
 
+		try {
+		    $newAvatar = $this->getAttributeValue('saml-attribute-mapping-avatar_mapping', $attributes);
+            $this->logger->debug('Avatar attribute content: {avatar}', ['app' => 'user_saml', 'avatar' => $newAvatar]);
+        } catch (\InvalidArgumentException $e) {
+            $this->logger->debug('Failed to fetch avatar attribute: {exception}', ['app' => 'user_saml', 'exception' => $e->getMessage()]);
+            $newAvatar = null;
+        }
```
blizzz, line 670: indentation

lib/UserBackend.php
```diff
@@ -677,9 +710,51 @@ public function updateAttributes($uid,
...
 				}
 			}
+
+			if ($newAvatar !== null) {
+                $image = new \OCP\Image();
+                $fileData = file_get_contents($newAvatar);
+                $image->loadFromData($fileData);
+                $this->setOwnCloudAvatar($uid, $image);
+            }
 		}
```
blizzz, line 720: indentation

lib/UserBackend.php
```diff
@@ -677,9 +710,51 @@ public function updateAttributes($uid,
...
+            return false;
+        }
+
+        $fs = new FilesystemHelper();
+        if (!$fs->isLoaded()) {
+            $fs->setup($uid);
+        }
+
+        try {
+            $avatarManager = \OC::$server->getAvatarManager();
```
blizzz, line 743: should be injected into the constructor instead

lib/UserBackend.php
```diff
@@ -677,9 +710,51 @@ public function updateAttributes($uid,
...
+			if ($newAvatar !== null) {
+                $image = new \OCP\Image();
+                $fileData = file_get_contents($newAvatar);
+                $image->loadFromData($fileData);
+                $this->setOwnCloudAvatar($uid, $image);
+            }
 		}
 	}
 
+    private function setOwnCloudAvatar($uid, $image) {
```
blizzz, line 723: that's copied from somewhere ;)

lib/UserData.php
```diff
@@ -37,11 +38,14 @@ class UserData {
...
 	/** @var IConfig */
 	private $config;
+	/** @var Image */
+	private $image;
 
 	public function __construct(UserResolver $userResolver, SAMLSettings $samlSettings, IConfig $config) {
 		$this->userResolver = $userResolver;
 		$this->samlSettings = $samlSettings;
 		$this->config = $config;
+		$this->image = new Image();
```
blizzz, line 48: it should be injected as parameter

tests/unit/Settings/AdminTest.php
```diff
@@ -139,6 +139,11 @@ public function formDataProvider() {
 				'type' => 'line',
 				'required' => true,
 			],
+            'avatar_mapping' => [
+                'text' => $this->l10n->t('Attribute to map the users avatar to.'),
+                'type' => 'line',
+                'required' => true,
+            ],
```
blizzz, line 146: indentation

lib/UserBackend.php
```diff
@@ -677,9 +710,51 @@ public function updateAttributes($uid,
...
+        $fs = new FilesystemHelper();
+        if (!$fs->isLoaded()) {
+            $fs->setup($uid);
+        }
+
+        try {
+            $avatarManager = \OC::$server->getAvatarManager();
+
+            $avatar = $avatarManager->getAvatar($uid);
+            $avatar->set($image);
```
blizzz, line 746: this writes the avatar every time. Especially when auth is happening often, this is a waste of resources and expensive (since it triggers several updates in other components).

blizzz (changes requested):

lib/UserBackend.php
Lumrenion: Yes,I refactor that if statement! The method evaluates if a user can change their avatar manually via the Nextcloud UI. If the argument is not existent in the user_saml configuration it is the same as being empty and the user can change their avatar. So the catch should be fine I think.

lib/UserBackend.php
Lumrenion: That is right. The avatar is updated every time any other SAML attribute is updated. I think this behaviour should be the same as in the LDAP extension. I agree that it might create unnecessary overhead. Do you have a suggestion how to improve this?

blizzz: In LDAP we store the and compare the hash of the image data. Only if it does not match, the received image data is set as new avatar.

it might be useful to move this into the avatar api to avoid duplication, but now this would earliest go into NC 23.

Lumrenion: can we use duplicate code here with an annotation to refactor it for NC23 (if the code moves into the avatar api)?

Lumrenion: I changed the code to save the checksum of the avatar image in user data and compare it before any image processing is done, when the user logs in.

RailForWork:
Love this function, when will it be commit?

blizzz:
I missed the notification in August. Will revisit after #558  is merged and shipped.

RailForWork:
> I missed the notification in August. Will revisit after #558 is merged and shipped.

Cool, thanks for your working.

blizzz:
@Lumrenion could you rebase, please?

blizzz:
Not sure the rebase worked with the 40 commits added here

Lumrenion:
True, it has commits on my fork that definitely came from this master. Should I create a new fork and manually do the changes again?

blizzz:
> True, it has commits on my fork that definitely came from this master. Should I create a new fork and manually do the changes again?

either retry to rebase, or do a new PR. You can apply your commits, not need to redo everything by hand.

Lumrenion:
Created a new PR with the actual changes only: https://github.com/nextcloud/user_saml/pull/640

Issue #403, opened 2020-02-25
Allow setting of "retrieveParametersFromServer"


Quebrantos:
I have been setting up a SSO/SLO environment with other applications (WordPress, and Microsoft OWA) using Microsoft ADFS as iDP . I had a weird issue when logged in into more than one application (NextCloud included) and initiating the logout from one of the other applications. NextCloud SLO proccess broke the chain of SAML Logout Requests/Responses not finishing its' own logout and not allowing the iDP to logout the rest of applications.

I have been able to fix the issue editing the file /apps/user_saml/lib/Controller/SAMLController.php and setting requestParametersFromServer to true in the porcessSLO call at line 354
[...]
                                $targetUrl = $auth->processSLO($keepLocalSession, null, **true**, null, $stay);
[...]

https://github.com/onelogin/php-saml/issues/130

LukasReschke:
Asked in https://github.com/onelogin/php-saml/issues/466 if we can remove this setting completely.

PR #525, opened 2021-05-04
Allow setting of "retrieveParametersFromServer"

LukasReschke:
Some SAML servers require this type of decoding, otherwise the SLO request fails. Ideally the library would perform both verifications (https://github.com/onelogin/php-saml/issues/466), but it seems upstream doesn't want to perform this change.

Until we have considered a better solution for this, this adds a new checkbox that one can configure.

Ref https://github.com/nextcloud/user_saml/issues/403

Signed-off-by: Lukas Reschke <lukas@statuscode.ch>

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -419,9 +419,9 @@ class SAMLController extends Controller {
 			if ($isFromIDP) {
 				$keepLocalSession = true ; // do not let processSLO to delete the entire session. Let userSession->logout do the job
 				$targetUrl = $auth->processSLO(
-					$this->SAMLSettings->usesSloWebServerDecode(),
+					$keepLocalSession,
 					null,
-					false,
+					$this->SAMLSettings->usesSloWebServerDecode(),
 					null,
 					$stay
 				);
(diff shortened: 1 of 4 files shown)


LukasReschke:
@Quebrantos, if you still require this change a quick test by you would be awesome! :)

lib/Controller/SAMLController.php
```diff
@@ -418,8 +418,14 @@ public function singleLogoutService() {
 			$stay = true ; // $auth will return the redirect URL but won't perform the redirect himself
 			if ($isFromIDP) {
 				$keepLocalSession = true ; // do not let processSLO to delete the entire session. Let userSession->logout do the job
```
blizzz, line 420: Is it art or may we remove it?

LukasReschke: Wait. I did a mistake here.... :/

LukasReschke: I should have replaced the third parameter but replaced the first one 🙈 - :/ :/ :/ - That's what you get for copy-pasting from one system to another.

MorrisJobke (approved):
Tested and works fine 👍

lib/Controller/SAMLController.php
LukasReschke: Background: I am unable to test this locally since the IDPs supporting this all cost $$$. And when I copy-pasted it from the support ticket, I seem to have put it into the wrong line. Sorry, and good catch!

blizzz (approved):

blizzz:
backports?

LukasReschke:
😢

LukasReschke:
Backport to 3.3 at https://github.com/nextcloud/user_saml/pull/526/files

PR #539, opened 2021-07-19
Integrate with Gitpod

JonathanTreffler:
[Gitpod](https://www.gitpod.io/) is a container-based development platform. Gitpod provisions ready-to-code development environments in the cloud accessible through your browser and your local IDE.
It enables you to describe your dev environment as code and start configurable and fresh development environments for each new task entirely in the cloud.

This sets up this repository to be able to open it in Gitpod with an automatic dev server.

Normally I would not be open to rely on such a platform, but since Gitpod is open source and can be self hosted i think it is okay. If anybody has a competing platform I should include I would be glad to do that. This is not a gitpod commercial, I just want to make development easier :)


blizzz:
I am closing this for inactivity, can be re-opened any time of course.

PR #542, opened 2021-09-01
use system email address getter if available

blizzz:
* [ ] requires https://github.com/nextcloud/server/pull/28422

The default getter might return the primary email, here we need the system mail only.


artonge (approved):

juliushaertl (approved):

skjnldsv:
Damn, merged the wrong PR 😩

juliushaertl:
Still fine to have that merged already as it will only call the new method if available

PR #554, opened 2021-10-12
Use effective uid for autoprovisioning new users

juliusknorr:
Fixes an issue where newly provisioned users get created with the base64 encoded uid instead of the decoded one.

The second commit was found on the way and avoids trying to decode the uid again which is alreay done before through getEffectiveUid()

### DIFF
--- a/tests/unit/Controller/SAMLControllerTest.php
+++ b/tests/unit/Controller/SAMLControllerTest.php
@@ -252,6 +252,12 @@ class SAMLControllerTest extends TestCase  {
 			->willReturn('MyUid');
 		$this->userData
 			->expects($this->any())
+			->method('testEncodedObjectGUID')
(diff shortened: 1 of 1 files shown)


blizzz:
> Only with patching in an objectguid value to the saml attributes, as i couldn't find a way to let my ldap server return such an attribute.

It is an AD thing. You could arrange this with Samba4, which is compatible and provides it, too, but (at leased used to) be tricky to set up.

I tried Azure AD, too, but they don't send it in binary i believe.

juliushaertl:
> It is an AD thing. You could arrange this with Samba4, which is compatible and provides it, too, but (at leased used to) be tricky to set up.
> I tried Azure AD, too, but they don't send it in binary i believe.
> P.S.: But you can also let Azure return a fake value.

Good to know, thanks :+1:

Issue #428, opened 2020-06-04
Can't setup SAML authentication


fedecastelli:
Hello,

I'm trying to setup the `SAML Authentication` using the [Ianum Identity Provider](https://www.ianum.com).

I fill all the Identity Provider's information and I see the error message saying `Invalid Metadata`. If I click on `Download Metadata`, the server shows an error page.

Checking the [official documentation](https://apps.nextcloud.com/apps/user_saml) I noticed that the Identity Provider is not officially supported and tested but it should be supported since it implements the SAML 2.0 protocol.

### Server configuration

**Operating system:** Ubuntu 18.04.6 LTS

**Web server:** Apache

**Database:** MySQL

**PHP version:** 7.2.24

**Nextcloud version:** 18.0.5

**Updated from an older Nextcloud/ownCloud or fresh install:** Fresh install

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>

```
[index] Error: OneLogin\Saml2\Error: Invalid array settings: idp_entityId_not_found, idp_sso_not_found, idp_cert_or_fingerprint_not_found_and_required at <<closure>>
0. /var/www/nextcloud/public/apps/user_saml/lib/Controller/SAMLController.php line 222
   OneLogin\Saml2\Settings->__construct({strict: true,de ... }})
1. /var/www/nextcloud/public/lib/private/AppFramework/Http/Dispatcher.php line 170
   OCA\User_SAML\Controller\SAMLController->getMetadata(null)
2. /var/www/nextcloud/public/lib/private/AppFramework/Http/Dispatcher.php line 99
   OC\AppFramework\Http\Dispatcher->executeController(OCA\User_SAML\Co ... {}, “getMetadata”)
3. /var/www/nextcloud/public/lib/private/AppFramework/App.php line 125
   OC\AppFramework\Http\Dispatcher->dispatch(OCA\User_SAML\Co ... {}, “getMetadata”)
4. /var/www/nextcloud/public/lib/private/AppFramework/Routing/RouteActionHandler.php line 47
   OC\AppFramework\App::main(“OCA\\User_SAML\ ... r”, “getMetadata”, OC\AppFramework\ ... {}, {_route: “user_saml.SAML.getMetadata”})
5. <<closure>>
   OC\AppFramework\Routing\RouteActionHandler->__invoke({_route: “user_saml.SAML.getMetadata”})
6. /var/www/nextcloud/public/lib/private/Route/Router.php line 299
[...]

PR #558, opened 2021-11-09
Move SAML configurations to a table of their own

blizzz:
- adds user_saml_configurations table and migrates existing configuration
- Controller methods are added since appconfig endpoints cannot be used
  anymore. THIS IS A BREAKING CHANGE.
- Frontend code is adjusted to use new endpoints.
- security-sloWebServerDecode was changed from global to provider specific
  setting. It being global seemed to be unintended. 
  - Similar case for 'sp-name-id-format'

* [x] fix »Primary index name on "oc_user_saml_configurations" is too long.«
* [X] ~~set security-sloWebServerDecode to each configuration~~ keep as is,  the setting had only effect  on the first configuration. Doing anything else might result in different behaviour than before, this should be avoided.
* [x] adjust unit tests
* [x] adjust integration tests, too
* [x] added OCC commands
* [x] consider OCS endpoint for configurations as replacement for appconfig endpoint → OCC
* [x] smoke testing

---

## occ commands

* saml:config:create

reserves and returns an new ID for the new SAML configuration. Not necessary to use, but if useful if there is a unknown number of confiugrations.  Example:

```
$ php occ saml:config:create
12
```

* saml:config:delete <providerId>

Deletes the config identified by the provided ID. No output on success.

```
$ php occ saml:config:delete 12
```

* saml:config:get [-p|--providerId PROVIDERID] [--output [OUTPUT]]

Returns all configuraitons or the one optionally specified by the --providerId parameter. Support the usual output formats (plain, json or json_pretty as of now). Example:

```
$ php occ saml:config:get --providerId=11 --output=json_pretty

{
    "11": {
        "general-uid_mapping": "http:\/\/schemas.xmlsoap.org\/ws\/2005\/05\/identity\/claims\/name",
        "general-idp0_display_name": "Byzantia II",
        "idp-entityId": "https:\/\/sts.windows.net\/0c4f0808-1234-abcd-9db8-1234abcdef56\/",
        "idp-singleSignOnService.url": "https:\/\/login.microsoftonline.com\/0c4f0808-1234-abcd-9db8-1234abcdef56\/saml2",
        "idp-singleLogoutService.url": "https:\/\/login.microsoftonline.com\/common\/wsfederation?wa=wsignout1.0",
        "idp-x509cert": "-----BEGIN CERTIFICATE-----\nMIIC8DCCAdigAwIBAgIQHH1B50eN+K1H4MDKJGt34jANBgkqhkiG9w0BAQsFADA0MTIwMAYDVQQD\nEylNaWNyb3NvZnQgQXp1cmUgRmVkZXJhdGVkIFNTTyBDZXJ0aWZpY2F0ZTAeFw0yMDExMDUxNDE5...\/iS\n-----END CERTIFICATE-----",
        "security-sloWebServerDecode": "0"
    }
}
```

* saml:config:set [options] [--] <providerId>

Sets the specified configuration parameters for the given SAML provider. See `--help` for a complete list of configuration parameters. On success, the output is empty. Example:

```
$ php occ saml:config:set --general-idp0_display_name="My favorite IdP"  --security-lowercaseUrlencoding=1  11
```


blizzz:
yeehaw, integration tests giving green lights :)

appinfo/routes.php
```diff
@@ -69,14 +69,32 @@
...
+			'url' => '/settings/providers',
+			'verb' => 'GET',
+		],
 		[
 			'name' => 'Settings#getSamlProviderSettings',
 			'url' => '/settings/providerSettings/{providerId}',
 			'verb' => 'GET',
 			'defaults' => [
-				'providerId' => '1'
+				'providerId' => 1
```
nickvergessen, line 82: add requirement for `\d+`

skjnldsv: Done

skjnldsv (approved):

skjnldsv:
Tests are failing on clone :thinking:

js/admin.js
```diff
@@ -200,7 +235,7 @@ $(function() {
 				});
 			});
 			$('input:checkbox[value="1"]').attr('checked', true);
-			$('input:checkbox[value="0"]').attr('checked', false);
+			$('input:checkbox[value="0"]').removeAttr('checked', false);
```
PVince81, line 238: `.prop('checked', false)` should work

js/admin.js
```diff
@@ -60,23 +70,49 @@
...
+			xhr.onerror = function () {
+				console.error("Could not fetch new provider ID");
+			};
+
+			xhr.send();
+		},
+
+		updateProvider: function (configKey, configValue, successCb, errorCb) {
+			var xhr = new XMLHttpRequest();
+			xhr.open('POST', OC.generateUrl('/apps/user_saml/settings/providerSettings/' + this.currentConfig));
```
PVince81, line 98: somehow it feels like POST and PUT were mixed up

usually in REST we use POST for adding new entries and PUT for updating one

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## 5.0.0
+### Changed
+- store configurations in a separate database table, not appconfig
+
+### Added
+- occ commands for modifying SAML configurations
```
PVince81, line 9: add section "Removed" with "ability to change SAML configuration with occ app-config, use the new occ commands instead"

this would highlight the breaking change for those using the old commands in scripting

lib/Command/ConfigSet.php
```diff
@@ -0,0 +1,84 @@
...
+			if(!in_array($key, SAMLSettings::IDP_CONFIG_KEYS) || $value === null) {
+				continue;
+			}
+			if ($value === '') {
+				unset($settings[$key]);
+				continue;
+			}
+			$settings[$key] = $value;
+		}
+		$this->samlSettings->set($pId, $settings);
```
PVince81, line 80: ideally for tools like Ansible to detect if a value was actually changed, we should output a message "value was changed" or "value was not changed"

lib/Command/ConfigDelete.php
```diff
@@ -0,0 +1,59 @@
...
+		$this->addArgument(
+			'providerId',
+			InputArgument::REQUIRED,
+			'ProviderID of the SAML config to edit'
+		);
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$pId = (int)$input->getArgument('providerId');
+		$this->samlSettings->delete($pId);
```
PVince81, line 56: should we output a message when config was deleted ? and when it did not exist ?

lib/Controller/SettingsController.php
```diff
@@ -78,43 +91,34 @@ public function getSamlProviderSettings($providerId) {
...
 				if (strpos($category, 'attribute-mapping') === 0) {
 					$category = 'attribute-mapping';
-					$key = $prefix . 'saml-attribute-mapping' . '-' . $setting;
+					$key = 'saml-attribute-mapping' . '-' . $setting;
+				} else {
+					$key = $category . '-' . $setting;
+				}
+
+				if (isset ($details['global']) && $details['global']) {
+					$settings[$category][$setting] = $this->config->getAppValue('user_saml', $key, '');
```
PVince81, line 100: I'm guessing this exists only for migration purposes, maybe add a PHP doc entry to clarify this

js/admin.js
```diff
@@ -8,20 +8,30 @@
...
+			xhr.open('GET', OC.generateUrl('/apps/user_saml/settings/providers'));
+			xhr.setRequestHeader('Content-Type', 'application/json');
+			xhr.setRequestHeader('requesttoken', OC.requestToken);
+
+			xhr.onload = function () {
+				var response = JSON.parse(xhr.response);
+				if (xhr.status >= 200 && xhr.status < 300) {
+					if (response.providerIds !== "") {
+						OCA.User_SAML.Admin.providerIds += ',' + response.providerIds;
+						OCA.User_SAML.Admin.currentConfig = OCA.User_SAML.Admin.providerIds.split(',').sort()[0];
```
PVince81, line 22: sorting might not be required, it seems the backend already does it: https://github.com/nextcloud/user_saml/pull/558/files#diff-7466884a39494cea3f50d3aaa2c0e3aed4adefb1630992261d6d56e6b812f8fbR56

(sorting could be slower for a lot of entries, but probably not a big issue)

lib/Db/ConfigurationsMapper.php
```diff
@@ -0,0 +1,102 @@
...
+	public function reserveId(): int {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('id')
+			->from('user_saml_configurations')
+			->orderBy('id', 'DESC')
+			->setMaxResults(1);
+
+		try {
+			$entity = $this->findEntity($qb);
+			$newId = $entity->getId() + 1;
```
PVince81, line 91: is this trying to anticipate autoincrement ? not sure if this works with all DBs

maybe add a PHPDoc to explain why this is needed

lib/Migration/Version5000Date20211025124248.php
```diff
@@ -0,0 +1,218 @@
...
+	public function __construct(IDBConnection $dbc) {
+		$this->dbc = $dbc;
+	}
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 */
+	public function preSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
```
PVince81, line 63: remove empty function ?

lib/Migration/Version5000Date20211025124248.php
```diff
@@ -0,0 +1,218 @@
...
+
+		$deletedRows = $deleteQuery
+			->setParameter('cfgKeys', $keys, IQueryBuilder::PARAM_STR_ARRAY)
+			->executeStatement();
+
+		return $deletedRows > 0;
+	}
+
+	protected function insertConfiguration(int $id, array $configData): bool {
+		static $insertQuery;
```
PVince81, line 147: should this be a class attribute ?

lib/Migration/Version5000Date20211025124248.php
```diff
@@ -0,0 +1,218 @@
...
+
+	protected function readConfiguration(array $configKeys): \Generator {
+		static $readQuery;
+		if (!$readQuery) {
+			$readQuery = $this->dbc->getQueryBuilder();
+		}
+		$readQuery->select('configkey', 'configvalue')
+			->from('appconfig')
+			->where($readQuery->expr()->eq('appid', $readQuery->createNamedParameter('user_saml')))
+			->andWhere($readQuery->expr()->in('configkey', $readQuery->createParameter('cfgKeys')));
```
PVince81, line 175: assuming less than 1000 entries, otherwise this needs chunking

lib/Migration/Version5000Date20211025124248.php
```diff
@@ -0,0 +1,218 @@
...
+use OCP\IDBConnection;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
+ */
+class Version5000Date20211025124248 extends SimpleMigrationStep {
+
+	private const IDP_CONFIG_KEYS = [
```
PVince81, line 20: dejavu from SAMLSettings, or are we duplicating these on purpose to not class-load the other class before running the migration ? (please answer in a PHPDoc that clarifies if applicable)

PVince81 (commented):
Apart from the POST/PUT potential mixup the rest is only minor comments.

Note: I only have very limited knowledge of this app and don't have an env to test it

lib/Migration/Version5000Date20211025124248.php
CarlSchwan: There is no way to hit the limitation since $configKeys is just the list of IDP_CONFIG_KEYS with a one prefix.

lib/Command/ConfigDelete.php
CarlSchwan: Also added a check to make sure that providerId is an int, otherwise a user could delete/change the providerId == 0 by mistake

lib/Migration/Version5000Date20211025124248.php
CarlSchwan: Also did that with readQuery and deleteQuery. I also noticed that readQuery wasn't correctly put inside the if statement

CarlSchwan:
> Tested locally and for me the migration wasn't run :/ I'll see if I'm able to find there the bug is

Found the issue: providerIds was not in the oc_appconfig table for me. The UI worked correctly for me since in that case getAppConfig for it used the default value of 1, but the migration didn't handle this case. Fixed now :)

lib/Migration/Version5000Date20211025124248.php
CarlSchwan: The two lists are not exactly the same. lib/SAMLSettings.php contains more entries. I'm not sure if this is expected or not. @blizzz

CarlSchwan (approved):
I looked into details in the code, fixed a few small issues and this looks fine to me now :)

CarlSchwan:
Found one more small issue: this PR change the min-version from 21 to 22 because of the executeQuery/executeStatement. This changes should be either reverted or indicated inside the Changelog/appinfo.xml file

blizzz:
> Found one more small issue: this PR change the min-version from 21 to 22 because of the executeQuery/executeStatement. This changes should be either reverted or indicated inside the Changelog/appinfo.xml file

Gna, forgot that it was not present in 21. We need this for 21, too, so should get back to the deprecated methods for now.

CarlSchwan:
> > Found one more small issue: this PR change the min-version from 21 to 22 because of the executeQuery/executeStatement. This changes should be either reverted or indicated inside the Changelog/appinfo.xml file
> 
> Gna, forgot that it was not present in 21. We need this for 21, too, so should get back to the deprecated methods for now.

Should be alright now :) I'll test the migration again to make sure nothing broke and then I think we can merge it

lib/Migration/Version5000Date20211025124248.php
```diff
@@ -192,17 +201,18 @@ protected function normalizeKey(string $prefixedKey): string {
 		throw new \RuntimeException('Invalid regex pattern');
 	}
 
+	/** @return list<int> */
```
blizzz, line 204: should be `int[]` or did i missed to learn a new trick?

blizzz:
Just stumbled over this artifact:

https://github.com/nextcloud/user_saml/blob/master/appinfo/app.php#L73-L87 especially line 76:
`$oneLoginSettings = new \OneLogin\Saml2\Settings($samlSettings->getOneLoginSettingsArray(1));`

Every request the first configuration is being read (since earlier than summer  2018) and loading tested. I interpret that code in the way that a configuration check is being done and fail gently on error. 

But the code was never adapted towards more configurations, and it also does not give a hint about potential issues. 

I am tempted to remove the whole try block.

lib/Migration/Version5000Date20211025124248.php
```diff
@@ -192,17 +201,18 @@ protected function normalizeKey(string $prefixedKey): string {
...
 			->from('appconfig')
 			->where($q->expr()->eq('appid', $q->createNamedParameter('user_saml')))
 			->andWhere($q->expr()->eq('configkey', $q->createNamedParameter('providerIds')));
 
-		$r = $q->executeQuery();
+		$r = $q->execute();
 		$prefixes = $r->fetchOne();
 		if ($prefixes === false) {
-			return [];
+			return [1]; // 1 is the default value for providerIds
```
blizzz, line 215: @CarlSchwan is it this part that broke for you when there was only one entry left, but  no  providerIds field present in the old appconfig? For when there is really no configuration it does not need to run at all. I would then add a check in `postSchemaChange()` to continue if `$configKeys` is empty and save readConfiguration (if only to ensure that in the DB query `cfgKeys` is not assigend an empty list and the engine may decide to error out – did not check it does, but not trying is not provoking).

lib/Migration/Version5000Date20211025124248.php
CarlSchwan: See https://psalm.dev/docs/annotating_code/type_syntax/array_types/#lists 

int[] is equivalent to `@psalm-return array<array-key, int>` while `list<int>` is a bit more exact and is `@psalm-return array<int, int>`

blizzz: tahnks, have not seen it  on  first  glance, i suspected it to be psalm. But my IDE was also confused ;) `@psalm-return` it should be then :)

lib/Migration/Version5000Date20211025124248.php
CarlSchwan: In https://github.com/nextcloud/user_saml/blob/d1f5a8314a618565237df98d1456f0ad0303165a/lib/SAMLSettings.php#L66 and https://github.com/nextcloud/user_saml/blob/d1f5a8314a618565237df98d1456f0ad0303165a/lib/Settings/Admin.php#L58 we are assuming that providerIds is by default `1`.

When I configured a SAML provider with master, my provider was properly configured in the UI and database but the providerIds key wasn't and because the default is 1, this worked (ok I was hit by https://github.com/nextcloud/user_saml/issues/428 but that's hopefully another issue :sweat_smile: ).

blizzz:
Identified another issue: when loading the SAML settings, the fields are empty. Only when you switch/click on a provider, the actual values are loaded.

And loading meta data for first provider (1) fails, because empty string is being sent up. Quick fix is easy, also looking for a better one tomorrow.

lib/Migration/Version5000Date20211025124248.php
blizzz: It  is sort of duplicate for we cannot rely on new classes on migration as old ones may have been autoloaded. Also, when new keys are added to SAMLSettings, they are not required in the Migration class for they cannot have been added in the past. Therefore this snapshot is also reasonable.

@CarlSchwan  thanks for pointing out the differences! They should be there indeed. 

There is also an existing issue with security-sloWebServerDecode which is IdP-sensitive but was always read from the first config.  So, tasks:

* [ ] fix security-sloWebServerDecode – set globally, but IdP sensitive
* [ ] migrate sp-x509cert, sp-name-id-format, sp-privateKey, security-sloWebServerDecode
* [ ] evaluate reading of sp-name-id-format in AdminSettings getForm (before, it took the first from config).

blizzz:
last checkboxes ticked, tomorrow for some testing and double checking that I did not oversee anything.

PVince81:
@blizzz I heard you had difficulties testing ? can you explain the troubles so perhaps someone can help out ?

PVince81:
as I understood this is blocked by https://github.com/nextcloud/user_saml/pull/583 ?
do we actually need that update to move forward or can the image update be done afterwards ?

tamahau30494:
Appreciate all the work done on this. I am wondering where this stands? I would like to be able to use the SAML plugin with version 23.

blizzz:
> Appreciate all the work done on this. I am wondering where this stands? I would like to be able to use the SAML plugin with version 23.

see the comment above…

PR #560, opened 2021-11-11
Update master target versions

nickvergessen:


PVince81:
I see there are integration tests for stable21 and stable22, maybe we should enable for stable23 as well ?
however the format fail, so probably something for later

PR #566, opened 2021-12-10
Apply triggers on all tests

skjnldsv:
Avoid having two different runs

image


skjnldsv:
Tests are failing because master is 24, fix here: https://github.com/nextcloud/user_saml/pull/567

PVince81 (approved):
:+1:

PR #576, opened 2022-01-21
Redirect to login page on sso login error

PVince81:


PVince81:
obsolete, need to use apache rules instead

Issue #80, opened 2017-01-20
Can't login "normally" with SSO enabled


r3pek:
While I have SSO enabled and working configured via server variables, I can't login normally to my next cloud instance if I'm on a computer that isn't in my Kerberos Domain.

Don't know if I'm missing something or not, but looking at the code at https://github.com/nextcloud/user_saml/blob/master/lib/Controller/SAMLController.php#L130-L135, instead of just showing a "Not provisioned" page, the "normal" login page should be shown (username/password login using whatever backend is cofigured)

LukasReschke:
At the moment that is the expected behaviour. I'm thinking about adding login via regular user account as well but this is not yet done.

jh23453:
I think right now nextcloud works a follows:

1. User calls /nextcloud/
2. The user is not authenticated to Nextcloud, so we'll redirect to /nextcloud/index.php/login
3. Nextcloud knows that we want to use SAML/SSO, we we'll redirect again, this time to  /nextcloud/index.php/apps/user_saml/saml/login. 
4. User get's authenticated with Kerberos or mod_auth_kerb can do basic auth (but no 2FA).
5. We'll redirect with authentication and the user can now work.

With SSO/Kerberos something like that might be more appropriate (see https://www.freeipa.org/page/Web_App_Authentication/Example_setup#Failed_Negotiate):

1. User calls /nextcloud/
2. Naxtcloud knows that we want to use SAML/SSO, we we'll redirect to 
/nextcloud/index.php/apps/user_saml/saml/login.
3.a. User get's authenticated with Kerberos (no fallback to basic auth).
3.b. The user is not authenticated to Nextcloud, so we'll redirect to /nextcloud/index.php/login for regular nextcloud login.
5. We'll redirect with authentication and the user can now work.

schiessle:
You could try to call `https://<your-nextcloud>?direct=1` This might forward you to the login page

PVince81:
> You could try to call `https://<your-nextcloud>?direct=1` This might forward you to the login page

this doesn't work for the login flow of the iOS app, it might require workarounds like suggested in https://github.com/nextcloud/user_saml/issues/400#issuecomment-751703259 to make it redirect to the "?direct=1" page

PVince81:
I was told that there used to be another button "alternative login", has this been removed ?

tomasdbs:
[...]

PR #577, opened 2022-01-21
Redirect to the password authentification page when kerberos is not available

CarlSchwan:
TODOS:

- [ ] It is a good idea?
- [ ] Should we add a config switch to opt-in or opt-out of this?


PVince81:
context: https://github.com/nextcloud/user_saml/issues/80

PVince81:
as discussed with @blizzz this is a rare scenario and adding this redirect here would confuse users

a redirection should be done in the web server directly

PR #579, opened 2022-01-28
Updating lint-php.yml workflow from template

nextcloud-command:
Automated update of the lint-php.yml workflow from https://github.com/nextcloud/.github


blizzz:
obsolote

PR #583, opened 2022-03-01
use PHP 8.0 image for master integration tests

blizzz:
* [x] depends on https://github.com/nextcloud/docker-ci/pull/359


blizzz:
* [ ] fix       cURL error 7: Failed to connect to ::1: Cannot assign requested address (see http://curl.haxx.se/libcurl/c/libcurl-errors.html) (GuzzleHttp\Exception\ConnectException)

blizzz:
Tests are still not stable, and **sometimes** fail with:

```
[18/Mar/2022:15:20:22.704103023 +0000] - ERR - add_new_slapd_process - Unable to start slapd because it is already running as process 9
[18/Mar/2022:15:20:22.705794092 +0000] - CRIT - main - Shutting down due to possible conflicts with other slapd processes
```

However there is nothing that would start the same thing twice. I even removed the start.sh as CMD from the dockerfile in https://github.com/nextcloud/docker-ci/pull/359, without effect. I thought for a weird reason it could be dependent on the host system (some workers are still on  Ubuntu 18.04), but when doing it manually i could not reproduce it. Nor can i locally.

PVince81:
@blizzz I searched online for that message, and found this: https://pagure.io/389-ds-base/issue/50989

are we stopping and restarting that service repeatedly ? maybe some lock cleanup missing ?

PVince81:
ok, so it seems there's a solution already: https://github.com/nextcloud/docker-ci/pull/359#discussion_r831022571

PVince81:
I've restarted the build and hope it will pick up the change there

artonge:
Solution seems to work :tada: 
Applying it to 8.0, rebuilding images

artonge:
All green, @blizzz DCO and fixup bot is failing

artonge (changes requested):
Remove vendor files?

blizzz:
> Remove vendor files?

sheesh, that opened another pandora's box

blizzz:
Mh, I am feeling to split the last commit from here and unifiy 3rdparty and vendor. Otherwise people will be mad to install it for it has to be composer install twice. This does not need to be another blocker.

CarlSchwan (approved):
:elephant:

PR #611, opened 2022-05-09
Cleanup app bootstrap

CarlSchwan:
Remove a bunch of deprecated stuff


lib/AppInfo/Application.php
```diff
@@ -21,14 +21,38 @@
 
 namespace OCA\User_SAML\AppInfo;
 
+use OC;
+use OC\Security\CSRF\CsrfTokenManager;
+use OC_User;
```
blizzz, line 26: private?

juliushaertl:
Just rebased, a few things left
- [x] Move script handling to event listener
- [x] Register middleware/service through register method
- [ ] Checkout if other private methods can be replaced
- [ ] registerDavAuth seems unused currently

come-nc:
Rebased and integrated commits on app.php since last time the PR was rebased.

come-nc:
Should be rebased after:

- [ ] https://github.com/nextcloud/user_saml/pull/942

To fix the psalm issue most likely.

Regarding `OC_User`, sadly it is necessary, until https://github.com/nextcloud/server/pull/52086 is merged. (But then it will still be necessary here unless you drop support for Nextcloud < 32)

Regarding other deprecated/private methods, I think cleaning that up should be postponed to another PR so that this one can be merged before conflicting again.

blizzz (approved):

Issue #573, opened 2022-01-11
Cant configure Saml settings, settings reset after reload/doesnt save


LittleITworld:
Hello,

Iam facing following issue:

Nextcloud version : **22.2.3**
Operating system and version: **Ubuntu 20.04**
Apache or nginx version:  **nginx/1.18.0**
PHP version: **8.0.13**

The issue you are facing:

* Cant configure SAML settings

If Iam trying to edit my settings its getting resetted everytime Iam reloading my browser (it says "saved") I receive the output from Admin Logging while entering my config (attached config 1) and my config looks like attached config 2 after refreshing the browser...

it also doesnt show me if my metadata are right or wrong, it happened after entering different urls in the saml settings. resetting saml (deactivate&remove-->install&activate) doesnt help aswell...

Is this the first time you've seen this error? _(Y/N)_: Y

Steps to replicate it:

1. install app
2. configure saml settings
3. refresh tab --> settings reset --> no SAML login button

The output of your Nextcloud log in **Admin > Logging**:
```
Error	index	OneLogin\Saml2\Error: Invalid array settings: idp_entityId_not_found, idp_sso_not_found		2022-01-10T11:43:16+0100
Error	index	OneLogin\Saml2\Error: Invalid array settings: idp_entityId_not_found, idp_sso_not_found		2022-01-10T11:43:15+0100
Error	index	OneLogin\Saml2\Error: Invalid array settings: idp_entityId_not_found, idp_sso_not_found, idp_cert_or_fingerprint_not_found_and_required		2022-01-10T11:43:03+0100
Error	index	OneLogin\Saml2\Error: Invalid array settings: idp_entityId_not_found, idp_sso_not_found, idp_cert_or_fingerprint_not_found_and_required		2022-01-10T11:42:47+0100
Error	index	OneLogin\Saml2\Error: Invalid array settings: idp_entityId_not_found, idp_sso_not_found, idp_cert_or_fingerprint_not_found_and_required		2022-01-10T11:42:33+0100
Error	index	OneLogin\Saml2\Error: Invalid array settings: idp_entityId_not_found, idp_sso_not_found, idp_cert_or_fingerprint_not_found_and_required		2022-01-10T11:42:19+0100
Error	index	OneLogin\Saml2\Error: Invalid array settings: idp_entityId_not_found, idp_sso_not_found		2022-01-10T11:42:09+0100
```

The output of your config.php file in `/path/to/nextcloud` (make sure you remove any identifiable information!):
```
[...]

PR #613, opened 2022-05-11
Fix displaying of global checkboxes in user_saml settings

KoffeinKaio:
Partly fix problems lined out in #573 
Templated wanted the values in config array, config array is only idp settings from user_saml, not the global ones.
Hope this approach is ok.

### DIFF
--- a/lib/Settings/Admin.php
+++ b/lib/Settings/Admin.php
@@ -186,10 +186,17 @@ class Admin implements ISettings {
 
 		$type = $this->config->getAppValue('user_saml', 'type');
 		if ($type === 'saml') {
+			$generalSettings['require_provisioned_account'] = [
+				'text' => $this->l10n->t('Only allow authentication if an account exists on some other backend (e.g. LDAP).', [$this->defaults->getName()]),
+				'type' => 'checkbox',
+				'global' => true,
+				'value' => $this->config->getAppValue('user_saml', 'general-require_provisioned_account', 0)
+			];
 			$generalSettings['use_saml_auth_for_desktop'] = [
 				'text' => $this->l10n->t('Use SAML auth for the %s desktop clients (requires user re-authentication)', [$this->defaults->getName()]),
 				'type' => 'checkbox',
 				'global' => true,
+				'value' => $this->config->getAppValue('user_saml', 'general-use_saml_auth_for_desktop', 0)
 			];
 			$generalSettings['idp0_display_name'] = [
 				'text' => $this->l10n->t('Optional display name of the identity provider (default: "SSO & SAML log in")'),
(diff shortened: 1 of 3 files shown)


lib/Settings/Admin.php
```diff
@@ -186,10 +186,17 @@ public function getForm() {
 
 		$type = $this->config->getAppValue('user_saml', 'type');
 		if ($type === 'saml') {
+			$generalSettings['require_provisioned_account'] = [
+                                'text' => $this->l10n->t('Only allow authentication if an account exists on some other backend (e.g. LDAP).', [$this->defaults->getName()]),
+                                'type' => 'checkbox',
+                                'global' => true,
+                                'value' => $this->config->getAppValue('user_saml', 'general-require_provisioned_account', 0)
+                        ];
```
CarlSchwan, line 194:
```suggestion
				'text' => $this->l10n->t('Only allow authentication if an account exists on some other backend (e.g. LDAP).', [$this->defaults->getName()]),
				type' => 'checkbox',
				'global' => true,
				'value' => $this->config->getAppValue('user_saml', 'general-require_provisioned_account', 0)
			];
```

tab vs space ;)

blizzz (changes requested):
there is something wrong with this PR.  Certainly the change in l10n should not be here.

KoffeinKaio:
ah heck, I made a mistake by rebasing. I instead merged all new commits into my repo. lemme think about how to fix this

lib/Settings/Admin.php
```diff
@@ -186,10 +186,17 @@ public function getForm() {
 
 		$type = $this->config->getAppValue('user_saml', 'type');
 		if ($type === 'saml') {
+			$generalSettings['require_provisioned_account'] = [
+				'text' => $this->l10n->t('Only allow authentication if an account exists on some other backend (e.g. LDAP).', [$this->defaults->getName()]),
+				type' => 'checkbox',
```
blizzz, line 191:
```suggestion
				'type' => 'checkbox',
```

KoffeinKaio:
aw to hell with that. lemme fix this again

blizzz:
@KoffeinKaio i took the freedom to rebase, drop the unneeded commits and amend the missing single quote fix to your main commit

KoffeinKaio:
> @KoffeinKaio i took the freedom to rebase, drop the unneeded commits and amend the missing single quote fix to your main commit

thanks, dont PR and work on other things my kids

KoffeinKaio:
nice one, just wanted to suggest adding another func to SAMLSettings as I'm not THAT familiar with PHPUnit. Will add that to my todo list.

blizzz:
> 



> nice one, just wanted to suggest adding another func to SAMLSettings as I'm not THAT familiar with PHPUnit. Will add that to my todo list.

these tests as they are written are not exactly elegant… but should work til next changes :D

blizzz (approved):

blizzz:
thanks for you contribution @KoffeinKaio :coffee:

PR #614, opened 2022-05-12
Revert "Handle mobile login flow with direct=1"

CarlSchwan:
This reverts commit 86684d6c5473765b9fc3c7f3146f214cbce0dfd4.

PR #617, opened 2022-05-24
5.0.1

blizzz:
* [ ] waits for #613

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ build
 .php_cs.cache
 
 !/tests/integration/
+
+.ackrc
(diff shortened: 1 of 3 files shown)


KoffeinKaio:
may I suggest doing the fix for https://github.com/nextcloud/user_saml/issues/428#issuecomment-1134550498 and including that in 5.0.1? 
Otherwise 5.0.1 will still be broken, although a bit prettier, but broken.

blizzz:
> may I suggest doing the fix for [#428 (comment)](https://github.com/nextcloud/user_saml/issues/428#issuecomment-1134550498) and including that in 5.0.1? Otherwise 5.0.1 will still be broken, a bit prettier, but broken.

It already has two other bugfixes (#612, #614) that are waiting for shipment. I like to get them out of the door without further ado. A 5.0.2 may follow up any time.

PR #647, opened 2022-08-24
fix signining in with IdPs other than 1

### LINKED ISSUE #630: Multiple IDP broken since 5.0.2
After updating to 5.0.2 only first idp is allways selected, selection of second, third... endpoints is not working. Most porobably this is in relation to the latest change, allways setting idp=1 by default

blizzz:
fixes #630 

The 'defaults' option in the route configuration is targetted at URL parameters only, but not GET parameters. The [dev docs](https://docs.nextcloud.com/server/latest/developer_manual/basics/routing.html) state:
> defaults (Optional): If this setting is given, a default value will be assumed for each URL parameter which is not present. The default values are passed in as a key => value par array

Therefore the defaults need to be written into the Controller method's signature.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -27,9 +27,6 @@ return [
 			'name' => 'SAML#login',
 			'url' => '/saml/login',
 			'verb' => 'GET',
-			'defaults' => [
-				'idp' => 1
-			],
 		],
 		[
 			'name' => 'SAML#base',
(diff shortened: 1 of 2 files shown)


blizzz:
CI failure unrelated (to be fixed with https://github.com/nextcloud/server/pull/33682)

CarlSchwan (approved):

PVince81 (approved):
:+1:

PR #648, opened 2022-08-24
(debug) output nextcloud log after integration tests ran

blizzz:
something was changed between Aug 6 and 24 that makes the integration tests fail…


blizzz:
ran them locally

PR #656, opened 2022-09-21
fix fetching meta data with IdP id 1

blizzz:
When IdP id is 1, the resulting URL is `https://example.org/index.php/apps/user_saml/saml/metadata?idp=` thus setting idp to effectively 0 by presence of the GET parameter, instead of defaulting to 1. Also we only need to provide the id without any other characters, simplifying this a bit.


juliushaertl (approved):
Code makes sense 👍

CarlSchwan (approved):

PR #545, opened 2021-09-07
Implementation of Groupbackend

JonathanTreffler:
continuation of #536

* [x]  Rebase to keep up with master
* [x] migration strategy
* [x] gid collision handling
* [x] automatically delete groups without members 
* [x] ensure admins cannot (un)assign members, create or delete saml groups
* [x] tests
* [x] Migrations for groups on existing installations

Since this is a long-running feature implementation with many PRs here's the timeline:
- #263
- #311
- #355
- #536
- this one

PR #659, opened 2022-09-23
Implement proper group mapping via SAML

### LINKED ISSUE #561: Grant Admin Rights to custom SAML group
## Is your feature request related to a problem? Please describe

When using SAML SSO, users must be added to the 'admin' group on the IdP as any changes made locally to the Nextcloud instance will be overwritten when the users logs in. There is no way to change or map a group to the 'admin' group in Nextcloud within the instance. This means the backend must have a group called 'admin' which does not clearly indicate what the group is for as it is likely the IdP is shared between multiple applications.

## Describe the behaviour you'd like

There should be in option in the SAML configuration to grant members of a Group admin rights to the Nextcloud application. This way the IdP could have a group such as *nextcloud-admins* be associated with members who should have admin rights to Nextcloud.

## Describe alternatives you've considered

Currently you can create a group within your IdP called 'admin'. For example, if using Keycloak you can create a local group called 'admin' and map the user to that group (you can do this with Roles as well).

## Additional context

This community topic also describes the issue
https://help.nextcloud.com/t/how-to-give-saml-users-admin-rights/115468

Ma27:
Fixes #561

As stated in the issue, it's not desirable to have a group called `admin` in the SAML backend which doesn't indicate to which service admin permissions are granted.

This is orthogonal to `saml-attribute-mapping-group_mapping` which simply maps all groups from a SAML attribute to Nextcloud groups, i.e. the attribute's value MUST contain a group called `admin` to make sure that users get admin rights in Nextcloud.

When enabled, the name of (another) attribute must be specified which contains a list of SAML-specific groups, e.g.

    ["nextcloud-admins", "nextcloud-marketing"]

that can be mapped to e.g.

    ["admin", "marketing"]

cc @jgallucci32, @kevinmccurdybrd, @blizzz

### DIFF
--- a/js/admin.js
+++ b/js/admin.js
@@ -223,6 +223,99 @@ $(function() {
 		OCA.User_SAML.Admin.resetSettings();
 	});
 
+	function isGroupMappingValid() {
+		if (!$('#user-saml-group-mapping-field-name').val()) {
+			return false;
+		}
+		var allMappingsDeclared = $('#saml-mapping-items')
+			.children('p')
+			.children('input')
+			.map(function () { return $(this).val(); })
+			.get()
+			.every(function (x) { return x; });
+		if (!allMappingsDeclared) {
+			return false;
+		}
+		return true;
+	}
+
+	function updateGroupMappings() {
+		OCA.User_SAML.Admin.setSamlConfigValue(
+			'group-mapping',
+			'field-mappings',
+			JSON.stringify(
+				$('#saml-mapping-items').children('p').map(function () {
+					return [$(this).children('.required').map(function () { return $(this).val() }).get()];
+				}).get()
+			)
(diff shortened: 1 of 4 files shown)


blizzz:
Did you see https://github.com/nextcloud/user_saml/pull/545 which is being in development?

Ma27:
OK interesting, this wasn't referenced in #561. 
What I'm wondering is: why do you have separate groups for SAML and non-SAML? IIRC most applications have a simple mapping between users from $directory and existing groups (or newly created ones) which is what my solution does (in a more simple fashion and without any migration steps).

Woudl be interested in knowing the use-case behind that, though :)

Ma27:
Closing due to lack of interest.

PR #670, opened 2022-11-16
add user filtering caps by group memberships

blizzz:
- allow to specify groups for rejection
- allow to specify groups required to auth

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -26,6 +26,7 @@ use Firebase\JWT\JWT;
 use OC\Core\Controller\ClientFlowLoginController;
 use OC\Core\Controller\ClientFlowLoginV2Controller;
 use OCA\User_SAML\Exceptions\NoUserFoundException;
+use OCA\User_SAML\Exceptions\UserFilterViolationException;
 use OCA\User_SAML\SAMLSettings;
 use OCA\User_SAML\UserBackend;
 use OCA\User_SAML\UserData;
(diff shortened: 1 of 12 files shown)


lib/Controller/SAMLController.php
```diff
@@ -157,6 +161,27 @@ private function autoprovisionIfPossible() {
...
+
+		$rejectGroupsString = $settings['saml-user-filter-reject_groups'] ?? '';
+		$rejectGroups = array_map('trim', explode(',', $rejectGroupsString));
+
+		if (!empty(array_intersect($groups, $rejectGroups))) {
+			throw new UserFilterViolationException('User is member of a rejection group.');
+		}
+
+		$requireGroupsString = trim($settings['saml-user-filter-require_groups'] ?? '');
+		$requireGroups = array_map('trim', explode(',', $requireGroupsString));
```
come-nc, line 179: It is not possible to use json because the UI is entering a comma separated group list, right?
At least avoid trimming twice.

```suggestion
		$requireGroupsString = $settings['saml-user-filter-require_groups'] ?? '';
		$requireGroups = array_map('trim', explode(',', $requireGroupsString));
```

come-nc (approved):

lib/Controller/SAMLController.php
blizzz: > It is not possible to use json because the UI is entering a comma separated group list, right?

that's correct. This trim is later for the `empty` check.

PR #672, opened 2022-11-18
user filtering only applicable with autoprov

blizzz:
When a provisioned account is required, it relies on another backend. Thus, allowing to auth depends on the presence of the user, and the enabled-state of the user.

Hence the configuration controls in the web UI are only shown when auto-provisioning is enabled. This change only implements a check and test cases in the business logic.

Follow up to #670


come-nc (approved):
:elephant:

PR #673, opened 2022-11-20
l10n: Add spaces

Valdnet:
Signed-off-by: Valdnet <47037905+Valdnet@users.noreply.github.com>


blizzz:
Should there be a space? Looks strange to me personally, but it might be just me.

PR #677, opened 2022-12-10
Use session locking to be compatible with Nextcloud 25 during logout

juliushaertl:
Hotfix to fix https://github.com/nextcloud/server/issues/35708 but makes sense anyways to have consistent session locking for the full logout method.

https://github.com/nextcloud/server/pull/35723 could be considered hardening for the server method so this PR allows fixing the issue separtely without requiring the server PR.


PVince81 (approved):
:+1:

come-nc (approved):

PVince81:
failing test unrelated, it also fails on master

PR #683, opened 2022-12-19
gently handle incoming SAML Logout Request when the session is missing

blizzz:
When the session, and thus the stored IdP ID, is not present, walk the available configurations, until one is found that does not result in errors. So far I did not take the permanent cookie route, as we cannot rely on it either and would need a fallback. It would be a good addition still. 

This might take a little when there are hundreds of SAML configurations though. Might need to have a better solution for those cases.

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -462,27 +462,23 @@ class SAMLController extends Controller {
 
 		if ($pass) {
 			$idp = $this->session->get('user_saml.Idp');
-			$auth = new Auth($this->samlSettings->getOneLoginSettingsArray($idp ?? 1));
-			$stay = true ; // $auth will return the redirect URL but won't perform the redirect himself
+			$stay = true; // $auth will return the redirect URL but won't perform the redirect himself
 			if ($isFromIDP) {
-				$keepLocalSession = true ; // do not let processSLO to delete the entire session. Let userSession->logout do the job
-				$targetUrl = $auth->processSLO(
-					$keepLocalSession,
-					null,
-					$this->samlSettings->usesSloWebServerDecode($idp),
-					null,
-					$stay
-				);
-
-				$errors = $auth->getErrors();
-				if (!empty($errors)) {
-					foreach ($errors as $error) {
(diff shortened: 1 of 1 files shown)


juliushaertl:
Seems reasonable. As far as I see it now we probably don't need a persistent cookie anyways but can just make sure to set the idp value in the session again after clearing it with the UserSession::logout method. That way we should not iterate over the providers in the 99% case I think.

Small patch that could achieve this but haven't tested yet:

```
diff --git a/lib/Controller/SAMLController.php b/lib/Controller/SAMLController.php
index 1393a15..39a7f25 100644
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -464,6 +464,7 @@ class SAMLController extends Controller {
                        $stay = true; // $auth will return the redirect URL but won't perform the redirect himself
                        if ($isFromIDP) {
+                               $this->session->remove('user_saml.Idp');
                                [$targetUrl, $auth] = $this->tryProcessSLOResponse($idp);
                                if ($auth !== null) {
                                        $errors = $auth->getErrors();
@@ -494,6 +495,9 @@ class SAMLController extends Controller {
                        if (!empty($targetUrl) && !$auth->getLastErrorReason()) {
                                $this->userSession->logout();
                        }
+
+                       // Add the IDP to the session again after logout to properly process the SLO response from the IDP
+                       $this->session->set('user_saml.Idp', $idp);
                }
                if (empty($targetUrl)) {
                        $targetUrl = $this->urlGenerator->getAbsoluteURL('/');
```

juliushaertl (approved):

blizzz:
The added remove should not be necessary, as the user logout process called later on would delete the session nevertheless. Having it set afterwards might work, but I doubt this is the case here. I suspect that the session was expired already, or the cookies were deleted. For, in every other case, this value would be present already. But i may oversee something?

juliushaertl:
The logout method would clear the session which generates a new session id, so this should lead to a new cookie, so from my perspective clearing and setting afterwards should work fine.

blizzz:
Yes, no doubt about this. Also a good point. Let me turn it into a new ticket, for I am off until second week in Jan.

PR #702, opened 2023-02-27
Remove useless base endpoint

provokateurin:
Not used anywhere and if a user would ever reach this route they would see a 404 anyway.


come-nc:
I’m not familiar enough with user_saml, @blizzz please review when you can

blizzz:
At least incomplete, for the route is at least also referenced here:

https://github.com/nextcloud/user_saml/blob/master/lib/SAMLSettings.php#L152

It was introduced inhttps://github.com/nextcloud/user_saml/pull/236 but I am also unsure about the intentions here.

PR #704, opened 2023-03-03
chore(CI): Adjust testing matrix for Nextcloud 26 on master

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

PR #705, opened 2023-03-03
Bump to v5.2 and require at least NC 25

blizzz:
Needed for https://github.com/nextcloud/user_saml/pull/698 as bumping php-jwt would not be backwards compatible with older NC/GSS versions.

Additional stable26 testing added in #704


come-nc (approved):

come-nc:
Codesniffer and drone are not happy.

blizzz:
> Codesniffer and drone are not happy.

Drone is actually happy, but 26 and master fail because of incompatibilites with the OS of the test image. There'S #712  on going to fix this, but we're not there yet. 25 is the canaray for so long :(

blizzz:
cannot reproduce the cs issue locally :-/

blizzz:
> cannot reproduce the cs issue locally :-/

eh, can, with PHP7.4 instead of 8 :exploding_head:

juliushaertl (approved):

PR #717, opened 2023-05-09
Avoid infinite redirection for disabled user

come-nc:
When a user is logged in through SAML while being disabled, they end up in an infinite redirection loop. This fixes it.


ArtificialOwl (approved):

come-nc:
https://github.com/nextcloud/user_saml/pull/705 is needed to fix CI

blizzz:
@come-nc is this needed for NC < 25?

Issue #694, opened 2023-01-24
Error Invalid argument supplied for foreach() at /var/www/nextcloud/apps/user_saml/lib/SAMLSettings.php#118 on config initialization


pboguslawski:
### Steps to reproduce
1. Put user_saml source in nc (tested v25.0.3) installation.
2. Configure `occ config:app:set user_saml general-require_provisioned_account --value='1'`
3. Configure `occ config:app:set user_saml type --value='environment-variable'`
4. Enable app with `occ app:enable user_saml`
5. Configure app with `occ saml:config:set --general-uid_mapping 'HTTP_X_FORWARDED_LOGIN' 1`

This will give us

```
# mysql -e 'select * from oc_appconfig' collab | grep saml
| user_saml               | enabled                               | yes 
| user_saml               | general-require_provisioned_account   | 1
| user_saml               | installed_version                     | 5.1.2
| user_saml               | type                                  | environment-variable|
| user_saml               | types                                 | authentication
```
saml auth works ok.

### Expected behaviour
No error like below in nc log. Option `Only allow authentication if an account exists on some other backend (e.g. LDAP).` checked in saml settings in nc UI.

### Actual behaviour
Nc log a few error messages with
```
Invalid argument supplied for foreach() at /var/www/nextcloud/apps/user_saml/lib/SAMLSettings.php#118
```
option `Only allow authentication if an account exists on some other backend (e.g. LDAP).` unchecked in saml settings in UI.

### Server configuration

**Operating system**: Debian 11

**Web server:** Apache

**Database:** MariaDB from Debian11

**PHP version:** PHP 7.4 from Debian 11
 
**Nextcloud version:** 25.0.3

**Where did you install Nextcloud from:** git source

**List of activated apps:**

irrelevant

**Nextcloud configuration:**

irrelevant

### Client configuration
**Browser:** irrelevant

**Operating system:** irrelevant

### Logs

As as above.

PR #720, opened 2023-05-12
Pass an array of Idp on the foreach loop

solracsf:
Fix https://github.com/nextcloud/user_saml/issues/694

Ensure type safety by making sure that the `$this->configurations` variable is properly initialized and contains an array before the foreach loop is executed.


solracsf:
Tests failing because of https://github.com/nextcloud/user_saml/pull/721

blizzz (changes requested):
instead, initialize `private $configurations;` as empty array, will save this condition

PR #721, opened 2023-05-12
Fix tests

solracsf:
26 is failing on SQLite due to https://github.com/nextcloud/server/issues/37649

### DIFF
--- a/.github/workflows/composer.yml
+++ b/.github/workflows/composer.yml
@@ -17,9 +17,9 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: install dependencies
-      uses: php-actions/composer@v5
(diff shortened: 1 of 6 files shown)


solracsf:
`integration-tests-master` outputs :

```
php /var/www/html/occ maintenance:install --database sqlite --admin-pass password
--
175 | Error while trying to initialise the database: An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1 near "(": syntax error
```

Isn't it the same error? 🤔

blizzz:
The underlying host OS is too old and some libs are incompatible

juliushaertl:
Just seen after i made tests passing again now after #732 so let me close this, thanks for the pr anyways.

PR #747, opened 2023-07-14
ci: Add psalm

juliushaertl:
👆 see title


blizzz:
the cs output is strange. It puts the opening curly bracket always into a new line – we never did that.

blizzz (approved):

PR #751, opened 2023-07-17
build(dependabot): commit vendor files in 3rdparty

blizzz:


juliushaertl:
Seems to fail https://github.com/nextcloud/user_saml/pull/747/checks?check_run_id=15109733726

As far as I see vendor is not available for composer: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#vendor

PR #754, opened 2023-07-21
Fix validating SAML XML schemas

nickvergessen:


juliushaertl:
- [ ] We'll need a follow up issue to see how we proceed with the custom patched 3rdparty

juliushaertl (approved):
Tested and works

nickvergessen:
We can also move the patch outside, and wrap each 3rdparty interaction instead

blizzz (changes requested):
let's not patch third party, packaging might override it

blizzz:
@nickvergessen @juliushaertl did the adjustments, and it is working on my test setup. If I get green lights from you, I'd clean up the commits before merging.

juliushaertl (approved):
Works here as well and code looks good 👍

patschi:
FWIW, I was experiencing same issues and driving me insane. Narrowed it down to XML validation today morning too. Applying this patch fixes it in my affected setup with ADFS too.

PR #764, opened 2023-08-09
migrate away from removed event dispatcher

icewind1991:
The event dispatcher is dead, long live the event dispatcher

PR #767, opened 2023-09-06
feat(server): master works now with upcoming NC 28

blizzz:
bump is needed for https://github.com/nextcloud/user_saml/pull/764


juliushaertl (approved):
Failures seem expected due to missing #764

blizzz:
So much to clean up now. Rabbit holes everywhere.

blizzz:
> Failures seem expected due to missing #764

Took a bit different approaches and removed usage of deprecated `IAppContainer` in `Application` and went for public `\OCP\Server` for other on-the-fly injections. In one class i could just use the `IEventDispatcher` that was actually already passed in the constructor.

blizzz:
That's just leaves the stub to be taken over from or in #764 actually

PR #769, opened 2023-10-06
fix(login): do not hardcode IdP to 1 on redirect

blizzz:
Given: the only available SAML configuration has an ID > 1 and we have a redirect scenario, the the SAML backend was hard requiring the config with IdP 1. This fixed with this PR by using the first (and only) available identifier.


susnux (approved):

juliushaertl (approved):

juliushaertl:
Do we need backports?

PR #771, opened 2023-10-10
fix: Implement IGetDisplayNameBackend

juliushaertl:
Otherwise instanceof checks like https://github.com/nextcloud/server/pull/38298/files#diff-2e1b245993621bf5efdce546bacce8f0e43efb7bc5d0c8d1b8a55f17231c4158R79 would fail

### DIFF
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -36,10 +36,11 @@ use OCP\IUserBackend;
 use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\Server;
+use OCP\User\Backend\IGetDisplayNameBackend;
 use OCP\User\Events\UserChangedEvent;
 use OCP\UserInterface;
 
-class UserBackend implements IApacheBackend, UserInterface, IUserBackend {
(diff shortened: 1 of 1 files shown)


lib/UserBackend.php
```diff
@@ -328,7 +331,7 @@ public function getDisplayName($uid) {
 			}
 		}
 
-		return false;
+		return $uid;
```
blizzz, line 334: did you check internal usage of this method? are we somewhere expecting or acting specifically to `false`?

juliushaertl: All other user backends implement it that way. The interface actually requires a string type and doesn't allow boolean values anymore

blizzz: No doubt about this, my point was to double check we don't have odd internal behavior that relies on the false.

blizzz: … but that is not the case :heavy_check_mark:

blizzz (approved):

blizzz:
Added a fix to another startic analysis finding https://github.com/nextcloud/user_saml/actions/runs/6473643774/job/17576817710?pr=771

blizzz:
@juliushaertl do we need a backport?

PR #774, opened 2023-10-26
fix: Do not try to validate metadata for environment variable config

juliushaertl:
There is no need to test metadata in this case.


datenangebot (approved):
Looks reasonable, even if I can not estimate all side effects

PR #797, opened 2023-12-20
fix: Also create skeleton for users from environment based auth

juliushaertl:
[Same logic as the regular SAML login](https://github.com/nextcloud/user_saml/blob/8a6a667e1496ec55dbfb92bb86817b7f2f396909/lib/Controller/SAMLController.php#L409) which should also apply when using environment based authentication

- [x] Needs to be tested


marcelklehr (approved):

marcelklehr:
thank you!

PR #815, opened 2024-03-07
feat(idps): Allow to filter idps based on trusted domains

jlehtoranta:
A Nextcloud installation might be available via multiple trusted domains (e.g. nextcloud.a.com and nextcloud.b.com). In this case an admin might want to link idps with specific trusted domains. For example associating only idp.a.com with nextcloud.a.com would redirect automatically to the idp login page, while at the same time there could still be multiple idps available via nextcloud.b.com.


blizzz:
Hi @jlehtoranta that looks good on first sight! Would it be possible for you to extend the feature with integration tests?

Issue #47, opened 2016-11-12
Enhance SP settings configuration


SalimHaddou:
Hi,

When I did my initial configuration I used the IP address of my Nextcloud, therefore the SP's metadata were not using the target SP domain name, so we should start the SAML config after adding a trusted domain name and managing nextcloud using it (that is not super obvious at first). Maybe you could add a field for the SP setting where the SP entity ID is configurable ?

Additionally, I am using Nextcloud behind a reverse proxy that does the SSL offloading. In this scenario, although the client is using HTTPS only, the AuthnRequest generated by the server contains a Issuer with http only (because it's hit using plain http on the serverside I suppose). It'd be great to support this scenario and have the SP aware it's being SSL offloaded by being able to configure (advanced view?) EntityID, Issuer, ACS... 

The first field under general is where the attribute is expected to match the username.
When a value is in that field, you no longer know what that field is for (the only explanation is there when the field is cleared). I normally use the SAML subject for user identification but here only an attribute works; couldn't you allow the subject to map the username and only use attributes for group, group admin, quota ?

Finally, do you intend to : 
Allow IdP metadata import to simplify IdP configuration ?
Support Multiple IdP (hostname based) ?

Thank you.
And keep up the good work !

Salim

solracsf:
Fixed some of the issues with https://github.com/nextcloud/user_saml/pull/816

blizzz:
- Multiple IdPs are long time there already, but not connected to hostnames. There is a contribution though that addresses it: https://github.com/nextcloud/user_saml/pull/815

- Config by meta data import is not there and currently it is not on the roadmap.

PR #816, opened 2024-03-11
fix(UI): permanent labels for input elements

blizzz:
Fixes the issue, reported also in #47, that means of input element fields where not clear once filled out, since only placeholders were used. I added labels now, and where applicable put example values as placeholders instead.

| Before | After |
|---|---|
| user_saml_before | user_saml_after |

### DIFF
--- a/templates/admin.php
+++ b/templates/admin.php
@@ -90,7 +90,8 @@ if (isset($_['general']['allow_multiple_user_back_ends']['text'])) {
 					</p>
 				<?php elseif ($attribute['type'] === 'line' && !isset($attribute['global'])): ?>
 					<p>
-						<input data-key="<?php p($key)?>" name="<?php p($key) ?>" value="<?php p($_['config']['general-'.$key] ?? '') ?>" type="text" <?php if (isset($attribute['required']) && $attribute['required'] === true): ?>class="required"<?php endif;?> placeholder="<?php p($attribute['text']) ?>"/>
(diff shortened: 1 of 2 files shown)


templates/admin.php
```diff
@@ -90,7 +90,8 @@
 					</p>
 				<?php elseif ($attribute['type'] === 'line' && !isset($attribute['global'])): ?>
 					<p>
-						<input data-key="<?php p($key)?>" name="<?php p($key) ?>" value="<?php p($_['config']['general-'.$key] ?? '') ?>" type="text" <?php if (isset($attribute['required']) && $attribute['required'] === true): ?>class="required"<?php endif;?> placeholder="<?php p($attribute['text']) ?>"/>
+						<label for="<?php p($key)?>"><?php p($attribute['text']) ?></label><br/>
+						<input data-key="<?php p($key)?>" name="<?php p($key) ?>" value="<?php p($_['config']['general-'.$key] ?? '') ?>" type="text" <?php if (isset($attribute['required']) && $attribute['required'] === true): ?>class="required"<?php endif;?> />
```
juliushaertl, line 94:
```suggestion
						<input data-key="<?php p($key)?>" id="<?php p($key) ?>" name="<?php p($key) ?>"  value="<?php p($_['config']['general-'.$key] ?? '') ?>" type="text" <?php if (isset($attribute['required']) && $attribute['required'] === true): ?>class="required"<?php endif;?> />
```

The label needs to point to an `id` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#for

juliushaertl (approved):

PR #819, opened 2024-03-14
fix(User): load timezone handling related resources

blizzz:
- the old style event name does not work here anymore


blizzz:
Test fails, because output format was changed in https://github.com/nextcloud/server/pull/39669 – ' instead of ` :see_no_evil:

juliushaertl (approved):

Issue #826, opened 2024-03-21
Call to undefined method OCA\User_SAML\UserBackend::checkPassword()


blizzz:
Well now I have another error, this time with impact :

[webdav] Erreur: Call to undefined method OCA\User_SAML\UserBackend::checkPassword()
	PROPFIND /remote.php/dav/files/admin/
	de 10.97.50.2 par -- à 21 mars 2024, 15:15:27

Impact is nextcloud windows client can no longer connect. I have the last stable version 3.12.2.

_Originally posted by @Alakadoo in https://github.com/nextcloud/user_saml/issues/825#issuecomment-2012459798_

blizzz:
@Alakadoo can you check your nextcloud.log for this error? Hopefully there is some backtrace included.

KnudH:
Same here

Complete Errormessage: 
``
[...]

Issue #825, opened 2024-03-21
nextcloud.oc_user_saml_auth_token missing errors in log since 6.1.2 update


Alakadoo:
### Steps to reproduce
1.Update user_saml to 6.1.2
2.Check logs

### Expected behaviour
No errors

### Actual behaviour
Lot of errors being logged, howerver no actual impact found

### Server configuration

**Operating system**: 
Ubuntu 22.04

**Web server:**
Apache 2.4.52

**Database:**
mysql 8.0.36

**PHP version:**
8.1.2

**Nextcloud version:** (see Nextcloud admin page)
28.0.3

**Where did you install Nextcloud from:**
Nextcloud server download

**List of activated apps:**

```
Enabled:
 - activity: 2.20.0
 - analytics: 4.12.0
 - bruteforcesettings: 2.8.0
 - calendar: 4.6.7
 - camerarawpreviews: 0.8.4
 - checksum: 1.2.3
 - circles: 28.0.0-dev
 - cloud_federation_api: 1.11.0
 - comments: 1.18.0
 - contacts: 5.5.3
 - dashboard: 7.8.0
 - dav: 1.29.1
 - deck: 1.12.2
 - dicomviewer: 2.0.0
 - federatedfilesharing: 1.18.0
 - federation: 1.18.0
 - files: 2.0.0
 - files_mindmap: 0.0.30
 - files_pdfviewer: 2.9.0
 - files_reminders: 1.1.0
 - files_sharing: 1.20.0
 - files_trashbin: 1.18.0
 - files_versions: 1.21.0
 - firstrunwizard: 2.17.0
 - forms: 4.1.1
 - groupfolders: 16.0.4
 - impersonate: 1.15.0
 - logreader: 2.13.0
 - lookup_server_connector: 1.16.0
 - nextcloud_announcements: 1.17.0
 - notifications: 2.16.0
 - notify_push: 0.6.10
 - oauth2: 1.16.3
 - officeonline: 2.1.0
 - password_policy: 1.18.0
 - photos: 2.4.0
 - privacy: 1.12.0
 - provisioning_api: 1.18.0
 - recognize: 6.1.1
 - recommendations: 2.0.0
 - related_resources: 1.3.0
 - serverinfo: 1.18.0
 - settings: 1.10.1
 - sharebymail: 1.18.0
 - spreed: 18.0.5
 - support: 1.11.0
 - survey_client: 1.16.0
 - systemtags: 1.18.0
[...]

PR #828, opened 2024-04-02
Restore checkPassword() function

solracsf:
Fix https://github.com/nextcloud/user_saml/issues/826
Regression from PR https://github.com/nextcloud/user_saml/pull/808

### DIFF
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -188,6 +188,34 @@ class UserBackend implements IApacheBackend, UserInterface, IUserBackend, IGetDi
 	}
 
 	/**
+	 * Check if the provided token is correct
+	 * @param string $uid The username
+	 * @param string $password The password
+	 * @return string
+	 *
+	 * Check if the password is correct without logging in the user
+	 * returns the user id or false
+	 */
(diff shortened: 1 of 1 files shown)


solracsf:
Seems the whole PR needs to be reverted instead...

jejbq:
Hello,

We are also affected by this issue #[825](https://github.com/nextcloud/user_saml/issues/825) #826 
so WebDAV does not work anymore ... 

If you cannot rollback PR #808, we will switch our Nextcloud AIO from SAML2 (user_saml) to OAuth2 (OIDC Connect)

Thank you in advance.

```
curl --user '*********:***************' -X PROPFIND -H "Depth: 1" 'https://nextcloud.example.com/remote.php/dav/files/username/'
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Error</s:exception>
  <s:message>Call to undefined method OCA\User_SAML\UserBackend::checkPassword()</s:message>
</d:error>
```

blizzz:
#829 is the a pull request with the fix…

PR #834, opened 2024-04-26
perf(db): Avoid double querying

ChristophWurst:
1. If you check for existence, then fetch a prop -> fetch the prop directly
2. If you check for existence, then delete -> delete directly

I have not tested these changes.


blizzz (approved):

blizzz:
now tested

blizzz:
@ChristophWurst do you fance a backport to cover NC 27 (and actually down to 25)?

PR #849, opened 2024-05-13
build(deps): update php-jwt to 6.10.0

blizzz:
We have to have the vendor files committed here at the moment – update after bump.


blizzz:
obsoloted by https://github.com/nextcloud/user_saml/pull/852

PR #851, opened 2024-05-13
Add some user related events

dermalikmann:
This issue initially popped up when looking into https://github.com/stjosh/auto_groups/issues/74

For now only added:

- `UserLoggedInEvent`
- `UserLoggedOutEvent`
- `UserDeletedEvent`

`UserCreatedEvent` should be added as well, but the ClassConstructor expects the users password, and I would need to investigate what possible options we have at this stage, as the user's real password is not known.

### DIFF
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -123,9 +123,9 @@ class UserBackend extends ABackend implements IApacheBackend, IUserBackend, IGet
 		if ($home !== '') {
 			//if attribute's value is an absolute path take this, otherwise append it to data dir
 			//check for / at the beginning or pattern c:\ resp. c:/
-				if ($home[0] !== '/'
-				   && !(strlen($home) > 3 && ctype_alpha($home[0])
-					   && $home[1] === ':' && ($home[2] === '\\' || $home[2] === '/'))
+			if ($home[0] !== '/'
+				&& !(strlen($home) > 3 && ctype_alpha($home[0])
+					&& $home[1] === ':' && ($home[2] === '\\' || $home[2] === '/'))
 			) {
 				$home = $this->config->getSystemValue('datadirectory',
 					\OC::$SERVERROOT.'/data') . '/' . $home;
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -103,41 +103,41 @@ class UserBackend extends ABackend implements IApacheBackend, IUserBackend, IGet
 	 * @param array $attributes
 	 */
 	public function createUserIfNotExists(string $uid, array $attributes = []): void {
-		if (!$this->userExistsInDatabase($uid)) {
-			$values = [
-				'uid' => $uid,
-			];
+		if ($this->userExistsInDatabase($uid)) return;
 
-			// Try to get the mapped home directory of the user
-			try {
-				$home = $this->getAttributeValue('saml-attribute-mapping-home_mapping', $attributes);
-			} catch (\InvalidArgumentException) {
-				$home = '';
-			}
+		$values = [
+			'uid' => $uid,
+		];
 
-			if ($home !== '') {
-				//if attribute's value is an absolute path take this, otherwise append it to data dir
-				//check for / at the beginning or pattern c:\ resp. c:/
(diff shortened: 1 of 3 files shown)


elainabialkowski:
Is there any update on this PR? I've noticed the checks having been sitting at Expected for a hot second. Is there anything I can do to help move this forward?

lib/UserBackend.php
```diff
@@ -118,41 +120,41 @@ protected function userExistsInDatabase(string $uid): bool {
...
-				'uid' => $uid,
-			];
-
-			// Try to get the mapped home directory of the user
-			try {
-				$home = $this->getAttributeValue('saml-attribute-mapping-home_mapping', $attributes);
-			} catch (\InvalidArgumentException) {
-				$home = '';
-			}
+		if ($this->userExistsInDatabase($uid)) return;
```
blizzz, line 123:
```suggestion
		if ($this->userExistsInDatabase($uid)) {
			return;
		}
```

Returning early is fine, but we always use `{ }` blocks, even if there is just one line of code (defensive approach).

blizzz (commented):
Could we also get in a test? There was one good attempt there that probably can be taken over, with adjusted classes: https://github.com/nextcloud/user_saml/pull/870/files#diff-de828f9df04940c610cecf388767da55bd9e27446a31c25ea1766c3d796d098e

blizzz:
@dermalikmann apologies for the late response and review. Are you still up for this?

dermalikmann:
I just read through #870
i could try to implement a test, but just as the other person, im not too familiar with nextcloud dev, so i need to take a look first. but you can expect something in a couple of hours.

dermalikmann:
@blizzz I copied over the test @x7airworker did in their PR.
I wanted to test for the other events as well, but neither the logout nor the user delete procedures have any tests yet, and I am not familiar enough with phpunit and this addon to implement even stubs for those tests. sorry 😅

Feel free to merge, resolve the conflict by using the incoming changes (mine).

blizzz:
>  I wanted to test for the other events as well, but neither the logout nor the user delete procedures have any tests yet, and I am not familiar enough with phpunit and this addon to implement even stubs for those tests. sorry 😅

Yes, and it is a bit nightmarish. We have integration tests, but those treat it as a black box. We can implicitly assume that they are covered. While not asserted, the call is simple and there is no complex logic behind, so it should get through. Not as good as an explicit test, but acceptable.

Rebased, resolved the conflict, and added minor fixes and style corrections.

blizzz:
… oh but this tests keeps failing now, needs another closer look.

dermalikmann:
Ok, so I took another look at this, and it turns out, blindly copying someone else's code is not always the best idea. The test where I copied the check into (`testLoginWithEnvVariable`) is testing the 'environment-variable' branch of lib/SAMLController.php::login()
This execution-path does not pass through lib/SAMLController.php::assertionConsumerService(), which is where the login event gets dispatched, as it's the step in the SAML login process. (it either redirects to the original call URL, the home page, or an error page in case of a failed login).

The easiest solution to make the test pass, would be to chuck another event dispatch into the login() function [here](https://github.com/nextcloud/user_saml/blob/e1e993181205393fcbab05e57f7fd5e783f8f7d0/lib/Controller/SAMLController.php#L243) which would be also an awesome place to add a "UserFirstTimeLoggedInEvent", but for now I added only the normal UserLoggedInEvent.

blizzz:
> The easiest solution to make the test pass, would be to chuck another event dispatch into the login() function [here](https://github.com/nextcloud/user_saml/blob/e1e993181205393fcbab05e57f7fd5e783f8f7d0/lib/Controller/SAMLController.php#L243) which would be also an awesome place to add a "UserFirstTimeLoggedInEvent", but for now I added only the normal UserLoggedInEvent.

The first UserFirstTimeLoggedInEvent fits there, while the regular one should be outside of the condition of course, but i agree!

blizzz:
@dermalikmann any updates on this?

CarlSchwan:
A similar page has now been merged and I marked it as co-authored by you as I used your PR as inspiration. Thanks!

PR #855, opened 2024-05-31
feat(groups): add setting display name to group backend

iLinaza:
SAML's GroupBackend does not allow you to set the display name of the groups.

The Nextcloud interface, ISetDisplayNameBackend, is implemented, which will enable setting the group name to work with SAML.

Also, group searching (on GroupBackend, getGroups method) also searches in display names.

### DIFF
--- a/lib/GroupBackend.php
+++ b/lib/GroupBackend.php
@@ -81,7 +81,7 @@ class GroupBackend extends ABackend implements IAddToGroupBackend, ICountUsersBa
 			)));
 			$query->orWhere($query->expr()->iLike('displayname', $query->createNamedParameter(
 				'%' . $this->dbc->escapeLikeParameter($search) . '%'
-			))); 
+			)));
 		}
 
 		if ((int)$limit > 0) {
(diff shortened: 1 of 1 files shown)


blizzz:
Hi @iLinaza thank you for your contribution. I wonder about the backgrounds of your change though. The group backend is presenting the groups as received from the IdP. Why should we allow manual changes of the group display name?

iLinaza:
Hi!

Thank you for your reply. In my case what I need is to be able to set the display name of the groups. Solving it as an IdP mapping would be enough in my case and, as you say, is what would make sense.

Actually, I may be wrong, the Group Display Name cannot be set as a mapping. Is this correct?

I might try revising the PR to make the DisplayName set as from a mapping if that is the case.

blizzz:
There is no mapping for group display names, no. That would also be strange as the response data is user focused. It would work if the group data structure we receive would contain both an identifier and a display name for each group, and this would need handling in our backend, as currently only a string list is a expected (for simplicity and compatibility this should be kept). This would my favored solution.

iLinaza:
If I understand correctly, do you mean changing the structure of the data received from the IdP and use a single mapping for both id and names? If this is your idea, I would need a little more detailed explanation to understand how this could be implemented.

Or that, in addition to sending a list of group identifiers for the existing group id mapping (saml-attribute-mapping-group_mapping), a list of group names should also be sent, in the same order for a new display name mapping?

The reason I want users to be able to see the group display name as well is because the group name is sometimes visible to them: 
 - From the profile, where they see which groups they belong to.
 - From the file sharing, in which they may want to share with their group and for this it is necessary to know the group name (the displayName, since by group id the group cannot be identified in a simple way).

blizzz:
@iLinaza I rebased and added a little adjustment, otherwise I'd take it for now as is.

blizzz (approved):

Issue #853, opened 2024-05-21
Use IEventDispatcher in GroupManager.php


bdovaz:
`GroupManager` currently relies on a [private and deprecated API](https://github.com/nextcloud/server/blob/master/lib/private/Hooks/PublicEmitter.php#L30):

https://github.com/nextcloud/user_saml/blob/af8911cc72f4843dbcd2c906db04edb9dd3c98d0/lib/GroupManager.php#L145

We should use the `IEventDispatcher` pattern as Nextcloud does in its `IGroupManager` implementation:

https://github.com/nextcloud/server/blob/5a6e48e85034aa81862b1fcbd93cb4ce58165e72/lib/private/Group/Manager.php#L287

Events:

https://github.com/nextcloud/server/tree/5a6e48e85034aa81862b1fcbd93cb4ce58165e72/lib/public/Group/Events

This interface [exists since version 17](https://github.com/nextcloud/server/blob/5a6e48e85034aa81862b1fcbd93cb4ce58165e72/lib/public/EventDispatcher/IEventDispatcher.php#L32).

If we don't want to be a breaking change, as seen in the example above, it sends it both ways (with `PublicEmitter` and `IEventDispatcher`).

@blizzz would you accept a PR?

PR #856, opened 2024-05-31
feat: migrate from deprecated PublicEmitter to IEventDispatcher

bdovaz:
Fixes https://github.com/nextcloud/user_saml/issues/853


bdovaz:
@blizzz I have doubts with these lines:

https://github.com/nextcloud/user_saml/blob/edb21705e2963322d38e0323baccd9bad69e0976/tests/stub.phpstub#L29

Should I remove them? I don't really know what this file is for, I come from the .NET world.  😅

blizzz:
> @blizzz I have doubts with these lines:
> 
> https://github.com/nextcloud/user_saml/blob/edb21705e2963322d38e0323baccd9bad69e0976/tests/stub.phpstub#L29
> 
> Should I remove them? I don't really know what this file is for, I come from the .NET world. 😅

So that's for testing, and the `PublicEmitter` is a private interface from the server, so it was stubbed here. By dropping that interface from `lib/GroupManager.php` we may do not need the stub anymore, and the entry can be removed just as well.

bdovaz:
> > @blizzz I have doubts with these lines:
> > https://github.com/nextcloud/user_saml/blob/edb21705e2963322d38e0323baccd9bad69e0976/tests/stub.phpstub#L29
> > 
> > Should I remove them? I don't really know what this file is for, I come from the .NET world. 😅
> 
> So that's for testing, and the `PublicEmitter` is a private interface from the server, so it was stubbed here. By dropping that interface from `lib/GroupManager.php` we may do not need the stub anymore, and the entry can be removed just as well.

Removed! PR ready to review!

blizzz:
There is failing CI. I see this also in an unrelated PR, so it do not think it is related to your changes. Nevertheless, I want to have this sorted out first so we have a green CI by default :)

blizzz:
> There is failing CI. I see this also in an unrelated PR, so it do not think it is related to your changes. Nevertheless, I want to have this sorted out first so we have a green CI by default :)

It was a regression within server, but fixed now.

blizzz (approved):

blizzz:
Now unit tests, they need adjustements. The IEventDispatcher has to be passed here https://github.com/nextcloud/user_saml/blob/master/tests/unit/GroupManagerTest.php#L57-L58 and maybe we want to check that `dispatchTyped()` is called with the correct event class.

bdovaz:
@blizzz done! Please rerun, thanks!

Is there any way to run the tests locally so that I don't depend on you to approve the workflow execution? Because looking at this file it seems that it depends on the fact that they must be executed through the workflow:

https://github.com/nextcloud/user_saml/blob/1a180957787d140af00d861847cc159cd84c0439/tests/unit/bootstrap.php#L10

blizzz:
If you have user_saml in an apps folder in a server checkout running locally will work. `composer i && composer run test:unit` should be sufficient then.

bdovaz:
@blizzz now I hope that everything is corrected, locally at least everything works correctly.

My problem was that I did not understand well the PHPUnit API and expects method when we want to check multiple invocations with different parameters of a method.

In my case I have not used the "at" method that is used in "UserBackendTest.php" because it says that it is deprecated and will be removed in the next version of PHPUnit. I could correct the one in "UserBackendTest.php" but I don't want to do it in this PR to avoid touching files that I should not.

bdovaz:
@blizzz I don't understand the errors that appear in the MySQL workflow, if the tests are exactly the same and in other database providers they have worked correctly (SQLite and PostgreSQL) how is it possible that in MySQL not and with this error?

```
1) OCA\User_SAML\Tests\GroupManagerTest::testUpdateUserGroups
Creation of dynamic property OCA\User_SAML\Tests\GroupManagerTest::$eventDispatcher is deprecated

/home/runner/actions-runner/_work/user_saml/user_saml/apps/user_saml/tests/unit/GroupManagerTest.php:50
```

image

bdovaz:
@blizzz I have corrected it anyway, it is that the field at class level was missing. The strange thing as I say is that it fails in some workflows but not in others. Apart from the fact that it is a warning but in reality it considers it an error.

You can run the build again.

blizzz:
> @blizzz I have corrected it anyway, it is that the field at class level was missing. The strange thing as I say is that it fails in some workflows but not in others. Apart from the fact that it is a warning but in reality it considers it an error.
> 
> You can run the build again.

Thanks! Dynamic properties were deprecated in PHP 8.2, so earlier ones would pass.

bdovaz:
@blizzz have passed all the checks!

Merge whenever you want, I'm all set!

Thanks.

blizzz:
Great! Thank you @bdovaz !

Issue #426, opened 2020-05-26
UserLoggedInEvent nor PostLoginEvent is emitted for user logins


The0tter0ne:
### Steps to reproduce
1. Configure SAML via AzureAD
2. Configure auto_groups to assign groups automatically to every user
3. Login with account via AzureAD SSO

### Expected behaviour
User logs in to nextcloud, gets groups assigned automatically.

### Actual behaviour
User logs in successfully but no groups are assigned 

### Server configuration
**Operating system:** CentOS 8
**Web server:** Nginx
**Database:** PostgreSQL 12.3 
**PHP version:** 7.3.17 
**Nextcloud version:** 18.0.4
**Where did you install Nextcloud from:** Linuxserver/nextcloud via docker hub

**List of activated apps:**
[Nextcloud_apps.txt](https://github.com/nextcloud/user_saml/files/4685520/Nextcloud_apps.txt)

**Nextcloud configuration:**
[Nextcloud_configuration.txt](https://github.com/nextcloud/user_saml/files/4685518/Nextcloud_configuration.txt)

### Client configuration
**Browser:** Firefox 76.01 & Chrome 83.0.4103.61
**Operating system:** Windows 10

#### Nextcloud log (data/owncloud.log)
[nextcloud.log](https://github.com/nextcloud/user_saml/files/4685577/nextcloud.log)

### Related issue
I have created a relatable issue with the other app, after coming to the conclusion it's probably this app causing the bug I raised the issue here. See: https://github.com/stjosh/auto_groups/issues/42 

The other app checks for UserLoggedInEvent or PostLoginEvents however this app apparently does not emit it.

adamcorcutt:
I'm having the exact same issue, would like to see this working.

menace133780:
I'm facing the same problem on NC 19.0.3

PR #870, opened 2024-07-31
Implemented PostLoginEvent in SAMLController

x7airworker:
This will also fix https://github.com/nextcloud/user_saml/issues/426.

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -20,6 +20,7 @@ use OCA\User_SAML\UserData;
 use OCA\User_SAML\UserResolver;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
+use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IConfig;
 use OCP\IL10N;
 use OCP\IRequest;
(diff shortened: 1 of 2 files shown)


blizzz:
Tests need adjustments. 

Also I just got aware about https://github.com/nextcloud/user_saml/pull/851 with different event classes :confused:

x7airworker:
Because I am using a custom user backend, I have no need for the events mentioned in https://github.com/nextcloud/user_saml/pull/851 although I agree that they should be implemented. I'll be happy to adjust the test cases if you can point me in the right direction. I am not familiar with writing tests for Nextcloud Apps and especially testing event dispatching, but I'll try my best.

x7airworker:
@blizzz Could you take a look at my added test assertions?

blizzz:
@x7airworker sure, I try to do it this week. Btw, pushing commits do not trigger Github notifications, a comment like yours now is helping catching attention :)

blizzz:
> Tests need adjustments.
> 
> Also I just got aware about #851 with different event classes 😕

@x7airworker sorry for bringing this up late. I was checking with the server team, and indeed we have here a duplicated event for whatever reason. The goal is now set to sunset `PostLoginEvent` in favor of `UserLoggedInEvent`, which is also  consistent with the modern naming of events. Again apologies for the forth-and-back. 

Thus is is better to continue with #851 (i have to have a closer look at it still), but the tests would be relevant there just as well.

PR #886, opened 2024-09-17
fix: Log attribute updates

ChristophWurst:
This should help localize slowness of the attribute update process.


blizzz:
It is quite talkative on the standard case (no updates). Intentional?

ChristophWurst:
It was intentional to be able to see the time between the skipped update and any slow update. But you can probably tell regardless.

blizzz (approved):

PR #888, opened 2024-09-18
fix(Groups): drop groups with mixed users from transition list

blizzz:
this avoids testing one group all over again and again, which might be costly when large groups have to be tested (user facing request!).

### DIFF
--- a/lib/GroupManager.php
+++ b/lib/GroupManager.php
@@ -70,6 +70,9 @@ class GroupManager {
 	private function getGroupsToRemove(array $samlGroupNames, array $assignedGroups): array {
 		$groupsToRemove = [];
 		foreach ($assignedGroups as $group) {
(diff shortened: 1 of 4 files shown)


lib/GroupManager.php
```diff
@@ -70,6 +70,9 @@ public function __construct(
 	private function getGroupsToRemove(array $samlGroupNames, array $assignedGroups): array {
 		$groupsToRemove = [];
 		foreach ($assignedGroups as $group) {
+			if (in_array($group->getGID(), $samlGroupNames)) {
```
ChristophWurst, line 73:
```suggestion
			if (in_array($group->getGID(), $samlGroupNames, true)) {
```

ChristophWurst (approved):
:+1: :elephant:

hweihwang (approved):

PR #901, opened 2024-11-05
feat(PassthroughParameters): Make it possible to pass through parameters to the SAML library

mickenordin:
OneLogin/Auth->login can accept an array of parameters to be passed through to the IdP.

This patch set makes it possible to specify an array of parameters to be passed through to the IdP, in the admin settings of the app

PR #903, opened 2024-11-14
fix(settings): Fix settings name

mickenordin:
I did a bunch more testing and in #901 it seems I got the name of the config key wrong in some places.

This patch fixes the issue.

### DIFF
--- a/lib/SAMLSettings.php
+++ b/lib/SAMLSettings.php
@@ -134,7 +134,6 @@ class SAMLSettings {
 			'strict' => true,
 			'debug' => $this->config->getSystemValue('debug', false),
 			'baseurl' => $this->urlGenerator->linkToRouteAbsolute('user_saml.SAML.base'),
-			'idp-passthroughParameters' => $this->configurations[$idp]['idp-passthroughParameters'] ?? '',
 			'security' => [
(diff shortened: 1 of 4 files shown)


lib/SAMLSettings.php
```diff
@@ -167,6 +166,7 @@ public function getOneLoginSettingsArray(int $idp): array {
 					'url' => $this->configurations[$idp]['idp-singleSignOnService.url'] ?? '',
 				],
 				'x509cert' => $this->configurations[$idp]['idp-x509cert'] ?? '',
+				'passthroughParameters' => $this->configurations[$idp]['idp-passthroughParameters'] ?? '',
```
blizzz, line 169: is this now correct with this 'idp-' text prefix?

blizzz (commented):
one question

lib/SAMLSettings.php
mickenordin: I now have this code running in 8 differnt sites, all working, so I think we are good.

blizzz (approved):

lib/SAMLSettings.php
blizzz: Yes, all good, boils down to the fact that we have to convert it to a php-saml compatible array structure. Hence the toying with the prefix.

PR #905, opened 2024-11-18
fix(Controller): make redirectUrl optional

blizzz:
… so that calling it does not break when doing it manually and the redirect parameter is lost.

The empty string is also the fallback if it the redirect target was not specified in normal operations.


blizzz:
Failing tests are PHP 8.4 related and due to the SAML lib that uses deprecated implicit nullable types. I.e. not relevant to this fix.

solracsf (approved):
💣 🏋️

blizzz:
> 💣 🏋️

:question: I don't think I am getting it. Slim Pickens does the right thing and carries the bomb to hell?

PR #906, opened 2024-11-19
fix(GroupBackend): implement missing ISetDisplayNameBackend

blizzz:
for we document that the groups display name can be change, but implementing this interface is actually necessary.


blizzz:
actually, a dup of https://github.com/nextcloud/user_saml/pull/855

blizzz:
closing in favor of #855

PR #915, opened 2024-12-04
build(release): 6.4.1

blizzz:
It is quick after 6.4.0, but it is another fix that got ready just afterwards I released yesterday and we would like to have this shipped until tomorrow. There is no regression in 6.4.0 or anything like this.


blizzz:
(Rebased because life is too short to wait for github runners)

hweihwang (approved):

PR #916, opened 2024-12-05
fix(UI): set value to the require_provisioned_account option also in env mode

blizzz:
Fixes the problem, that the option always appeard as disabled on a page reload:

[Screencast_20241205_230835.webm](https://github.com/user-attachments/assets/b4edf03c-f60d-430c-a603-4e8c45e599b7)


juliusknorr (approved):

juliusknorr:
Unit test failures need an adjustment

PR #932, opened 2025-02-08
feat(settings): optional config option for sp entityId

sniegel-mind4bytes:
Right now the service provider entityId is always generated based on the nextcloud instance URL. At least Keycloak relies on that entityId to resolve the corresponding SAML2 client. This PR introduces a new optional config option to allow the admin to set a different service provider entityId manually. If no manual configuration is done, the current behavior remains as the default.

### DIFF
--- a/lib/SAMLSettings.php
+++ b/lib/SAMLSettings.php
@@ -59,7 +59,7 @@ class SAMLSettings {
 		'saml-attribute-mapping-group_mapping_prefix',
 		'saml-user-filter-reject_groups',
 		'saml-user-filter-require_groups',
-        'sp-entityId',
+		'sp-entityId',
 		'sp-x509cert',
 		'sp-name-id-format',
 		'sp-privateKey',
--- a/lib/SAMLSettings.php
+++ b/lib/SAMLSettings.php
@@ -59,6 +59,7 @@ class SAMLSettings {
 		'saml-attribute-mapping-group_mapping_prefix',
 		'saml-user-filter-reject_groups',
 		'saml-user-filter-require_groups',
+        'sp-entityId',
 		'sp-x509cert',
 		'sp-name-id-format',
 		'sp-privateKey',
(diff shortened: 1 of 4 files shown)


blizzz:
@sniegel-mind4bytes hey and thanks for your contribution! 

Could you elaborate where the current mechanism does not meet your needs?

sniegel-mind4bytes:
> @sniegel-mind4bytes hey and thanks for your contribution!
> 
> Could you elaborate where the current mechanism does not meet your needs?

For me to have the SAML authentication work with Keycloak I have to use https://example.com/apps/user_saml/saml/metadata as the clientId for the SAML client on Keycloak. Instead of using the URL as clientId I would like to use something like "nextcloud". The issue extends a bit further as I have to use a single Keycloak for multiple staging environments. Hence the clientId should be something like, "test-nextcloud" or "prod-nextcloud". Indeed that would work with the current implementation of user_saml as the URLs differ for each stage. The thing is, nextcloud is the only SAML client requiring the URL as clientId. All the others give me the flexibility to use whatever clientId I like. Therefore I thought it would be great to give admins the same flexibility with user_saml as well. For those with no need, everything stays the same, as it is no requirement to set this config option, making this change completely backward compatible.

sniegel-mind4bytes:
@blizzz: Hi, any update? Let me know in case I can do something to push this pr further.

blizzz:
Hey @sniegel-mind4bytes, thanks for your patience (been off before and had a lot going on then) and for your explanation. Makes sense, I concur!

blizzz:
@sniegel-mind4bytes when i set the entityid it works, but when i unset it again, I run into an internal server error with `Invalid array settings: sp_entityId_not_found`.  I suppose because the key is still present, alas empty. Could you double check that?

sniegel-mind4bytes:
Hi @blizzz , any update on this PR?

blizzz (approved):

blizzz:
@sniegel-mind4bytes Again apologies for the delay, and thank you for your contribution!

PR #943, opened 2025-04-15
fix: Register the dav plugin through an event listener

come-nc:
Sabre plugins from info.xml are loaded only after auth has taken place,
 this is too late for user_saml. So this goes back to using an event,
 but uses the modern typed version.


blizzz (approved):

come-nc:
https://github.com/nextcloud/server/actions/runs/14472621054/job/40592798421

Not fixed yet it seems.
Maybe we should add the same tests to the CI here.

come-nc:
It’s a mistery to me why we are not listening to SabrePluginAuthInitEvent instead, but as this is complicated enough, for now listening to the same event as before to make sure it does not break anything.
`SabrePluginAuthInitEvent` is not emitted in all the same places so I won’t take the risk to change.

come-nc:
https://github.com/nextcloud/server/actions/runs/14472621054/job/40594700851 now it works, good for merging.

blizzz:
Thanks @come-nc :bow:

PR #944, opened 2025-04-17
chore: Add vendor-bin dependabot configurations

come-nc:


come-nc:
There is no npm in this application, so only adding vendor-bin repos in the end.

susnux (approved):

PR #950, opened 2025-05-02
Drop support for NC 28+29, requires PHP >= 8.1, update deps, add rector

blizzz:
- NC 28 and 29 are EOL (stable-6  is branched off)
- Hence we can bump the PHP requirements
- And so also bump dependencies
- Sugar on top is the rector configuration


mejo- (approved):
Nice to see all the code cleanup changes from rector :rocket:

elzody (approved):

PR #953, opened 2025-05-11
chore(i18n): Changed spelling of entity ID

rakekniven:
Reported at Transifex


blizzz:
Could you also adjust the string in `tests/unit/Settings/AdminTest::formDataProvider()`? :bow:

PR #962, opened 2025-05-20
feat: add saml:user:add command to pre-provision users

### LINKED ISSUE #939: Add saml:user:add command to pre-provision users
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

**Which Nextcloud Version are you currently using:** 31.0.2

**Is your feature request related to a problem? Please describe.**
Our Nextcloud instance allow logins from various institutions with a SAML federation, but we want to control which users are authorized to connect so no autoprovision.

**Describe the solution you'd like**
We can script pre-provisionning provided there is a saml:user:add command

`php occ saml:user:add uid --email=new.user@example.org --display-name="new user"`

**Describe alternatives you've considered**
Didn't find any. Since institutions in the federation are different we cannot easily add an attribute for our need. We have only our own users in LDAP.

**Additional context**
I'm not a PHP developper but I've coded something that seems to work. I'm not familiar with nextcloud dev process so I will just post it below hoping someone could adapt it.

guerby:
feat: add saml:user:add command to pre-provision users, fix #939

### DIFF
--- a/lib/Command/UserAdd.php
+++ b/lib/Command/UserAdd.php
@@ -27,7 +27,7 @@ class UserAdd extends Base {
 	protected function configure(): void {
 		$this
 			->setName('saml:user:add')
-			->setDescription('adds an saml account')
+			->setDescription('Add a SAML account')
 			->addArgument(
 				'uid',
 				InputArgument::REQUIRED,


lib/Command/UserAdd.php
```diff
@@ -0,0 +1,87 @@
...
+		parent::__construct();
+	}
+	protected function configure(): void {
+		$this
+			->setName('saml:user:add')
+			->setDescription('adds an saml account')
+			->addArgument(
+				'uid',
+				InputArgument::REQUIRED,
+				'Account ID used to login (must only contain a-z, A-Z, 0-9, -, _ and @)'
```
blizzz, line 36:
```suggestion
				'Account ID as provided by the IdP (must only contain a-z, A-Z, 0-9, -, _ and @)'
```

lib/Command/UserAdd.php
```diff
@@ -0,0 +1,87 @@
...
+			)
+			->addOption(
+				'display-name',
+				null,
+				InputOption::VALUE_REQUIRED,
+				'Login used in the web UI (can contain any characters)'
```
blizzz, line 42:
```suggestion
				'Name as presented in the web interface (can contain any characters)'
```

lib/Command/UserAdd.php
```diff
@@ -0,0 +1,87 @@
...
+			)
+			->addOption(
+				'email',
+				null,
+				InputOption::VALUE_REQUIRED,
+				'When set, users may register using the default email verification workflow'
```
blizzz, line 48: Could you clarify what is meant by this?

lib/Command/UserAdd.php
```diff
@@ -0,0 +1,87 @@
...
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$uid = $input->getArgument('uid');
+		
+		if ($this->userManager->userExists($uid)) {
+			$output->writeln('<error>The account "' . $uid . '" already exists.</error>');
+			return 1;
+		}
+
+		if (!$output->isQuiet()) {
+			$output->writeln('<info>The account "' . $uid . '" is to be added to SAML.</info>');
```
blizzz, line 61:
```suggestion
			$output->writeln('<info>The account "' . $uid . '" is to be added to the SAML backend.</info>');
```

To avoid confusion that someone might understand it would write upstream.

blizzz (commented):
Thank you for your  contribution!

Some suggestions on user facing text mostly. Please update the commit message to be meaningful, e.g. how i edited the PRs title.

lib/Command/UserAdd.php
guerby: I copy/pasted this option including the doc string from https://github.com/nextcloud/server/blob/master/core/Command/User/Add.php#L79 since unfortunately I'm not familiar at all with nextcloud code base.

I think it is indeed not appropriate for user_saml because there's no email workflow in this way to create user. 

As far as I understand email is needed at least for some form of notifications by nextcloud (but may be other things I'm not aware of) and is user editable after initial setting (in our nextcloud instance at least).

I'm open to suggestion, maybe "Set user default email in user profile" ?

Thanks again for your review!

guerby: I have updated the PR with "Set user default email in user profile"

blizzz:
FWIW Github does not create notifications when a commit was added.

lib/Command/UserAdd.php
blizzz: It is already possible to create users without email addresses as admin via both web interface and occ.

guerby:
> FWIW Github does not create notifications when a commit was added.

I have commited the change to make email VALUE_OPTIONAL, I think this is the last pending review item, let me know if I missed something, thanks!

blizzz:
`composer i && composer run cs:fix` should help solve the last failing test.

Plusses, if the `UPDATE $filename` commits can be squashed to their parent :)

guerby:
> composer i && composer run cs:fix

On my debian 12 I did `apt-get install composer php-xml php-curl` and ran `composer i && composer run cs:fix`.

For the squash part I'm reading git documentation, I assume you mean commits 29ba79493d137cfe9b85009ef8c414fddc1a294d  2cced28ae65ba8fceff8fe11ceae52523c1aa5cf and b886e05177e6fe2171cc567ea5c78c6167c4c165 ?

blizzz:
> > composer i && composer run cs:fix
> 
> On my debian 12 I did `apt-get install composer php-xml php-curl` and ran `composer i && composer run cs:fix`.

Thanks!

> For the squash part I'm reading git documentation, I assume you mean commits [29ba794](https://github.com/nextcloud/user_saml/commit/29ba79493d137cfe9b85009ef8c414fddc1a294d) [2cced28](https://github.com/nextcloud/user_saml/commit/2cced28ae65ba8fceff8fe11ceae52523c1aa5cf) and [b886e05](https://github.com/nextcloud/user_saml/commit/b886e05177e6fe2171cc567ea5c78c6167c4c165) ?

Aye :slightly_smiling_face:

guerby:
> i think something went wrong, the changes in the `l10n` should not be here though (they will be imported automatically from the translation tool)

I did the following to get rid of the unwanted commits:

```
git rebase -i HEAD~20 # removed the l"picked" ines of all unwanted commits
git push origin patch-user-add-fix-939 --force-with-lease
```
They seem to have disappeared now.

guerby:
Hmm they seem to reappear immediately, may be some bot is doing something?

lib/Command/UserAdd.php
```diff
@@ -0,0 +1,88 @@
...
+		protected IUserManager $userManager,
+		protected UserBackend $backend,
+		private LoggerInterface $logger,
+	) {
+		parent::__construct();
+	}
+	protected function configure(): void {
+		$this
+			->setName('saml:user:add')
+			->setDescription('adds an saml account')
```
ZephOne, line 30:
```suggestion
			->setDescription('Add a SAML account')
```

guerby:
Then I cliked on github web "merge" button

blizzz:
added a commit with just code style changes. it seems the cs check workflow also updates the dependencies, so that it runs php-cs-fixer in a different version than otherwise is installed.

PR #978, opened 2025-08-13
wip

kesselb:


lib/UserResolver.php
```diff
@@ -24,15 +24,25 @@ public function __construct(IUserManager $userManager) {
...
 	 */
 	public function findExistingUserId(string $rawUidCandidate, bool $force = false): string {
-		if ($force) {
-			$this->ensureUser($rawUidCandidate);
+		if ($force)
+		{
+			$usersFound=$this->ensureUser($rawUidCandidate);
+			if(count($usersFound) != 1)
+			{
+				throw new NoUserFoundException('User' . $rawUidCandidate . ' not valid or not found');
```
blizzz, line 32: This will work on maybe 99% of the cases, but not where the search attributes are configured in a way that would not reliably find the user.

Should not throw but continue with previous behaviour.

lib/UserResolver.php
```diff
@@ -24,15 +24,25 @@ public function __construct(IUserManager $userManager) {
...
+			}
+			$user = array_keys($usersFound)[0];
```
blizzz, line 34: There is a risk that more records are returned and that the first one is not the correct one.

It would be acceptable to cycle through the result and compare with the uid candidate (after sanitizing), for instance.

blizzz (changes requested):
The changes are not resulting in reliable behaviour in this form (plus code style).

PR #997, opened 2025-11-13
search disabled users by display name or email

### LINKED ISSUE #994: Sub-admins cannot search disabled SAML users by display name or email
### Steps to reproduce
1. Run Nextcloud 31.0.10 Enterprise with `user_saml` 7.0.0.
2. Create a sub-admin (delegated admin) who manages at least one group that comes from SAML provisioning.
3. Disable a user that belongs to that group.
4. Log in as the sub-admin, open Settings → Users, switch to “Disabled accounts”, and search for the user by display name or email.

### Expected behaviour
The disabled account appears when searching by display name or email, matching the behaviour full admins get.

### Actual behaviour
The search returns no results unless the sub-admin searches by the account name (UID). Full admins can find the same user via display name or email.

### Analysis
- Sub-admins hit `apps/provisioning_api/lib/Controller/UsersController.php:236`, which delegates to each managed group’s `searchUsers($search)` and filters disabled users.
- For SAML groups, `apps/user_saml/lib/GroupBackend.php:120` only applies the search term to `uid`. Display name and email are ignored, so the filtered list is empty unless the UID matches.
- Full admins bypass the group backend and use `lib/private/User/Manager.php:292::getDisabledUsers()`, where the search includes UID, display name, and email. That’s why they don’t experience the issue.

### Environment
- Nextcloud 31.0.10 Enterprise (`31.0.10.3`)
- user_saml 7.0.0
- PHP 8.2.29, Apache 2.4.62, MySQL 10.11.10
- Browser: Chrome 141 (Windows 10)

hweihwang:
https://github.com/nextcloud/user_saml/issues/994

### DIFF
--- a/lib/GroupBackend.php
+++ b/lib/GroupBackend.php
@@ -155,15 +155,35 @@ class GroupBackend extends ABackend implements IAddToGroupBackend, ICountUsersBa
 	#[\Override]
 	public function usersInGroup($gid, $search = '', $limit = -1, $offset = 0): array {
 		$query = $this->dbc->getQueryBuilder();
-		$query->select('uid')
-			->from(self::TABLE_MEMBERS)
-			->where($query->expr()->eq('gid', $query->createNamedParameter($gid)))
-			->orderBy('uid', 'ASC');
+		$query->selectDistinct('m.uid')
+			->from(self::TABLE_MEMBERS, 'm')
+			->where($query->expr()->eq('m.gid', $query->createNamedParameter($gid)))
+			->orderBy('m.uid', 'ASC');
 
 		if ($search !== '') {
-			$query->andWhere($query->expr()->like('uid', $query->createNamedParameter(
-				'%' . $this->dbc->escapeLikeParameter($search) . '%'
-			)));
+			$query->leftJoin('m', 'accounts_data', 'dn',
+				$query->expr()->andX(
+					$query->expr()->eq('dn.uid', 'm.uid'),
+					$query->expr()->eq('dn.name', $query->createNamedParameter('displayname'))
+				)
+			);
+			$query->leftJoin('m', 'accounts_data', 'em',
+				$query->expr()->andX(
+					$query->expr()->eq('em.uid', 'm.uid'),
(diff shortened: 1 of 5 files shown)


janepie (approved):

lib/GroupBackend.php
```diff
@@ -7,6 +7,7 @@
 
 namespace OCA\User_SAML;
 
+use OC\User\LazyUser;
```
blizzz: This is problematic, it is not part of the API. It is also bad that  the internal backend makes use of it, while there is no chance for external ones. 

Would it be sufficient to implement `IProvideEnabledStateBackend` instead? Then the user management abstraction would create the LazyUser objects.

lib/GroupBackend.php
```diff
@@ -154,15 +154,35 @@ public function groupExistsWithDifferentGid(string $samlGid): ?string {
...
+			);
+
+			$searchParam1 = $query->createNamedParameter('%' . $this->dbc->escapeLikeParameter($search) . '%');
+			$searchParam2 = $query->createNamedParameter('%' . $this->dbc->escapeLikeParameter($search) . '%');
+			$searchParam3 = $query->createNamedParameter('%' . $this->dbc->escapeLikeParameter($search) . '%');
+			$query->andWhere(
+				$query->expr()->orX(
+					$query->expr()->ilike('m.uid', $searchParam1),
+					$query->expr()->ilike('dn.value', $searchParam2),
+					$query->expr()->ilike('em.value', $searchParam3)
```
blizzz, line 184: having one $searchParam should be sufficient, as they are all the same.

blizzz (approved):

lib/GroupBackend.php
CarlSchwan: iirc this does not work with either oracle or PostgreSQL, I had unit tests fail in the past on one of them because I tried to be clever xD

CarlSchwan:
Need a rebase ;)

blizzz:
@hweihwang does it need a backport to stable-7 for 31 or 30?

PR #1003, opened 2025-12-08
refactor: Fix many psalm issues

CarlSchwan:


lib/Command/UserAdd.php
```diff
@@ -52,7 +52,8 @@ protected function configure(): void {
 	protected function execute(InputInterface $input, OutputInterface $output): int {
 		$uid = $input->getArgument('uid');
 
-		if ($this->userManager->userExists($uid)) {
+		$user = $this->userManager->get($uid);
+		if ($user === null) {
```
blizzz: Introduces a logic error. This must be:

```suggestion
		if ($user !== null) {
```

lib/Controller/TimezoneController.php
```diff
@@ -28,15 +30,18 @@ public function __construct(
...
 	 * @throws \OCP\PreConditionNotMetException
 	 * @throws \InvalidArgumentException
 	 */
+	#[NoAdminRequired]
+	#[UseSession]
 	public function setTimezone(string $timezone, int $timezoneOffset): JSONResponse {
 		if (!in_array($timezone, \DateTimeZone::listIdentifiers())) {
 			throw new \InvalidArgumentException('Invalid timezone');
 		}
+		if ($this->userId === null) {
```
blizzz, line 42: Psalm does not understand it cannot be null here

lib/Listener/SabrePluginEventListener.php
```diff
@@ -11,17 +11,34 @@
...
 			return;
 		}
-		$event->getServer()->addPlugin(Server::get(DavPlugin::class));
+		$event->getServer()->addPlugin(new DavPlugin(
+			$this->session,
+			$this->appConfig,
+			$_SERVER,
+			$this->settings,
+			$this->sessionService,
+		));
```
blizzz: should it not be creatable from `$event->getServer->get(…)`?

lib/Settings/Admin.php
```diff
@@ -91,7 +90,7 @@ public function getForm() {
 				'text' => $this->l10n->t('Only allow authentication if an account exists on some other backend (e.g. LDAP).', [$this->defaults->getName()]),
 				'type' => 'checkbox',
 				'global' => true,
-				'value' => $this->appConfig->getAppValueInt('general-require_provisioned_account'),
+				'value' => $this->appConfig->getAppValueBool('general-require_provisioned_account'),
```
blizzz: technically not a refactor, but a data type change. maybe it is ok. some with the change below.

lib/GroupManager.php
```diff
@@ -173,6 +185,9 @@ protected function createGroupInBackend(string $gid, ?string $originalGid = null
 		}
 
 		$group = $this->groupManager->get($gid);
+		if ($group === null) {
+			throw new GroupNotFoundException('Could not found group ' . $gid);
```
blizzz:
```suggestion
			throw new GroupNotFoundException('Could not find group ' . $gid);
```

lib/UserBackend.php
```diff
@@ -268,31 +243,17 @@ public function getDisplayNames($search = '', $limit = null, $offset = null): ar
...
-	 * In case the user has been authenticated by Apache true is returned.
-	 *
-	 * @return boolean whether Apache reports a user as currently logged in.
-	 * @since 6.0.0
-	 */
-	#[\Override]
+	#[Override]
 	public function isSessionActive(): bool {
-		return $this->session->get(SessionData::KEY_IDENTITY_PROVIDER_ID) !== null;
+		return $this->session->get('user_saml.samlUserData') !== null;
```
blizzz: Why is a different value used here?

lib/UserData.php
```diff
@@ -54,19 +54,26 @@ public function getEffectiveUid(): string {
...
 			return [];
 		}
 
-		return is_array($this->attributes[$mapping])
-			? $this->attributes[$mapping]
-			: array_map(trim(...), explode(',', (string)$this->attributes[$mapping]));
+		if (is_array($this->attributes[$mapping])) {
+			/** @var list<string> $groups */
+			$groups = $this->attributes[$mapping];
+			return $groups;
```
blizzz, line 73:
```suggestion
			return $this->attributes[$mapping];
```

blizzz (changes requested):

lib/Controller/TimezoneController.php
CarlSchwan: it's defined as `?string $userId` in the constructor

lib/Settings/Admin.php
CarlSchwan: right, to play it safe, I changed that back to int

lib/UserBackend.php
CarlSchwan: Fixed

lib/UserData.php
CarlSchwan: The temporary variable is to add an explicit type to make psalm happy

lib/Controller/TimezoneController.php
blizzz: Sure. Yet it is a controller and this is not a public page, hence we have an authed user.

lib/UserData.php
blizzz: :shrug:

blizzz (approved):
Nothing blocking left :)

lib/Controller/TimezoneController.php
CarlSchwan: But then if a non-authenticated user is trying to access the page, the error displayed won't be an authorization error from the  `#[NoAdminRequired]` attribute but an invalid argument exception

Ideally we could inject $userId into the controller method instead of the constructor of the Controller, then we would known that the user is logged in.

Issue #998, opened 2025-11-14
Logout option shown to other backend-provisioned users after session renewal (e.g. browser close)


blizzz:
When a SLO URL  is not set, the SAML backend does not offer the option to logout.

Providing a custom logout URL (or disabling it) is a user backend feature. As long as a user, that is provisioned through a different backend (e.g. LDAP) has the session after the SAML login, according to session data the SAML backend can enforce its logout URL policy.

After a browser restart however, this information is not present anymore, so the SAML backend does not kick into action and the default logout URL is being used.

blizzz:
I think `user_oidc` has a similar issue, reaching out to @julien-nc for a common understanding of a fix.

PR #1022, opened 2025-12-15
fix(Session): restore session data after cookie login

blizzz:
when a browser is closed and the session lost, a new session might be started with the help of the remember me cookie. For we keep some data in the session, like the IdP identifier, and some SAML data needed for SLO, those were lost in this process. This made especially logout behaviour not acting as expected.

Now we keep the required data also in the database and can restore it on a cookie login event.

Fixes https://github.com/nextcloud/user_saml/issues/998

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,11 +23,11 @@
 		"post-update-cmd": [
 			"[ $COMPOSER_DEV_MODE -eq 0 ] || composer bin all update --ansi"
 		],
-		"cs:fix": "php-cs-fixer fix",
-		"cs:check": "php-cs-fixer fix --dry-run --diff",
-		"psalm": "psalm",
-		"psalm:fix": "psalm --alter --issues=InvalidReturnType,InvalidNullableReturnType,MissingParamType,InvalidFalsableReturnType",
-		"psalm:update-baseline": "psalm --threads=1 --update-baseline",
+		"cs:fix": "@php php-cs-fixer fix",
+		"cs:check": "@php php-cs-fixer fix --dry-run --diff",
+		"psalm": "@php psalm",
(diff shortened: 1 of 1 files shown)


appinfo/info.xml
```diff
@@ -20,7 +20,7 @@ The following providers are supported and tested at the moment:
 	* Any other provider that authenticates using the environment variable
 
 While theoretically any other authentication provider implementing either one of those standards is compatible, we like to note that they are not part of any internal test matrix.]]></description>
-	<version>7.1.1</version>
+	<version>7.1.2-beta.1</version>
```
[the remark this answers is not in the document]
ChristophWurst: AFAIK there is no hard requirement on needing a major bump for db changes. Even patches can have them. Admins are not fans, but it's inevitable for some changes.

hweihwang (approved):

lib/Service/SessionService.php
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	public function storeIdentityProviderInSession(int $idp): void {
+		$this->session->set(SessionDataModel::KEY_IDENTITY_PROVIDER_ID, $idp);
+	}
+
+	public function restoreSession(string $oldSessionId): void {
+		try {
+			$sessionIdHash = $this->hashSessionId($oldSessionId);
+			$sessionData = $this->sessionDataMapper->retrieve($sessionIdHash);
```
kesselb: Should we store the user_id together with the session data as additional hardening? 

https://github.com/nextcloud/server/blob/ed6d0e59bb7726ecf2d9de14e76a8d24a65dd25b/lib/base.php#L1175-L1180 

Looks like we only reach this point when username + token + session_id matches?

blizzz: Since we restore the session when the successful-cookie-login-event has happened, this check was run already by the server and we can rely on it. No necessity to repeat the check here.

PR #1028, opened 2026-01-07
fix: DavPluging arguments

CarlSchwan:
Entity $id is now string|int|null in master so this is not needed and
even before this was not needed as typing on $id wasn't enforced.
    
The only important part are:
    
- `addType('id', STRING)` for the DB type to be correct
- `@method` tag in the api doc for psalm

### DIFF
--- a/lib/Db/SessionData.php
+++ b/lib/Db/SessionData.php
@@ -12,12 +12,12 @@ use OCP\AppFramework\Db\Entity;
 use OCP\DB\Types;
 
 /**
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

CarlSchwan:
> not complete, check #1027 instead

Done

> also not sure what the unrelated sessiondata changes are about

This is to make psalm happy, with the latest change in master. `$id` is now typed as `int|string|null`. I'll update this PR

PR #1035, opened 2026-01-27
drop support for NC 30 and 31; add support for upcoming NC 34

blizzz:
reducing the span of supported version a bit with adding 34 support

* [x] branch off stable-7 from master before merging

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -43,7 +43,7 @@ jobs:
       matrix:
         php-versions: [ "8.1", "8.2", "8.3", "8.4" ]
         databases: [ "mysql" ]
-        server-versions: [ "stable30", "stable31", "stable32", "stable33", "master" ]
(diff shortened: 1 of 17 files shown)


CarlSchwan:
If we can do a release of user_saml with https://github.com/nextcloud/user_saml/pull/1044 and NC 31 that would be quite helpful

blizzz:
> If we can do a release of user_saml with #1044 and NC 31 that would be quite helpful

Release is out, and this PR is rebased.

CarlSchwan (approved):

blizzz:
ok, something changed in server it seems, need to check it out

kesselb:
Hey @blizzz,

I'm going to pin server 31 and 31 to 7.1.4 for now to unblock the CI. Please branch off when time allows, then we can switch to a branch in the workflow.

blizzz:
Just branched-off stable-7 from v7.1.4

PR #1049, opened 2026-03-17
fix: Only set valid timezone values

come-nc:


AndyScherzinger (approved):

susnux (approved):

ernestofranze:
The fix from this PR was merged into `master` on Mar 17, 2026, but the backports to `stable31`, `stable32`, and `stable33` all failed (the branches don't exist on the remote).

The latest releases (v8.3.0, v8.2.1, v8.2.0) **do not include this fix** — `TimezoneController.php` still has the original buggy validation:

```php
if (!in_array($timezone, \DateTimeZone::listIdentifiers())) {
    throw new \InvalidArgumentException('Invalid timezone');
}
```

This causes `InvalidArgumentException('Invalid timezone')` errors (level 3) in the log every time a user accesses NC with a browser that sends deprecated timezone aliases like `America/Buenos_Aires` (Chrome on Windows does this via `Intl.DateTimeFormat().resolvedOptions().timeZone`).

Could this fix be included in the next release (v8.3.1 or v8.4.0)? Many installations are affected since `America/Buenos_Aires` and other deprecated aliases are still commonly sent by browsers.

blizzz:
Wrong PR? The lines you reflect are that ones introduces by this PR. This is part of 8.3.0, servicing NC 32 to 35.

ernestofranze:
You're right, my apologies for the confusion! I was looking at the code after upgrading to 8.3.0 and misidentified the fix as the bug. The `in_array` validation is indeed the fix introduced by this PR, and it IS present in 8.3.0.

The "Invalid timezone" errors we were seeing were the validation working as designed — rejecting deprecated aliases like `America/Buenos_Aires` that some browsers (Chrome on Windows) send via `Intl.DateTimeFormat().resolvedOptions().timeZone`.

We had a local patch that mapped deprecated aliases to canonical ones before the validation, but we've now removed it to keep the integrity check clean. The fix from this PR is working correctly in 8.3.0.

The remaining question is whether deprecated timezone aliases (that `DateTimeZone::listIdentifiers()` doesn't include but `Intl` API still returns) should be handled separately — but that's a different issue. Thanks for the quick response!

PR #1060, opened 2026-04-09
fix: Cleanup DI in Application::boot

come-nc:


CarlSchwan (approved):

lib/AppInfo/Application.php
```diff
@@ -82,24 +78,18 @@ public function boot(IBootContext $context): void {
...
 			): void {
-				$groupBackend = Server::get(GroupBackend::class);
-				Server::get(IGroupManager::class)->addBackend($groupBackend);
-
-				$userBackend = Server::get(UserBackend::class);
+				$groupManager->addBackend($groupBackend);
 
 				$userBackend->registerBackends($userManager->getBackends());
-				OC_User::useBackend($userBackend);
+				$userManager->registerBackend($userBackend);
```
come-nc, line 92: This change does not work properly on versions older than 32, despite the function existing the backend will not be known to OC_User and that causes troubles.

CarlSchwan: See the change in https://github.com/nextcloud/user_saml/pull/1061/changes which I copied from user_odic

come-nc: See merged https://github.com/nextcloud/user_saml/pull/1035 :smile:

PR #1072, opened 2026-04-15
feat: Trigged UserLoggedInEvent when user log in

CarlSchwan:
Took over from #851

We don't need all the events from that merge requests as some of these are already triggered by other methods (e.g. logout event is triggered in `userSession->logout()`)


artonge (approved):

blizzz (approved):
Alternative would be to call `User\Session::completeLogin` which does the same, and also would trigger user home init on first login and such, could then be avoided here. Not sure why we do not have it, possibly because it is not exposed in an API.

PR #1078, opened 2026-04-20
fix(settings): Make sure global settings are saved

CarlSchwan:
Andd add debouncing when using env-variable type as the settings are then saved immediatly.


come-nc (approved):
Tested, works

PR #1091, opened 2026-04-30
fix: Avoid redirecting to login with an incomplete configuration

### LINKED ISSUE #840: Handle incomplete saml configuration more gracefully
### Steps to reproduce
1. Enable user_saml
2. Use built-in SAML authentication
3. Look around, do not configure saml properly
4. Go to sleep
5. Visit Nextcloud (the session expired, and I was redirected to the login page)
6. Greeted by an error message

Screenshot from 2024-05-09 15-41-47

### Expected behaviour

I think it would be possible to handle such an incomplete but enabled saml configuration more gracefully.

### Actual behaviour

An incomplete saml configuration leads to an error message on the login page.

Incomplete saml configuration => Layer 8 problem, although we could handle it better.

### Server configuration

**PHP version: 8.3**

**Nextcloud version: master/30** (see Nextcloud admin page)

**Where did you install Nextcloud from: git**

come-nc:
- Fix https://github.com/nextcloud/user_saml/issues/840

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -167,7 +167,13 @@ class Application extends App implements IBootstrap {
 				}
 
 				$multipleUserBackEnds = $samlSettings->allowMultipleUserBackEnds();
-				$configuredIdps = $samlSettings->getListOfIdps();
+				$configuredIdps = $samlSettings->getListOfIdps(true);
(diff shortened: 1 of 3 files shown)


CarlSchwan (approved):
psalm need a bit of fixing but looks good

PR #1103, opened 2026-05-07
feat: Flexible matching between LDAP and SAML user ids

CarlSchwan:
Test plan:

1. Setup LDAP with an user `uid=bender mail=bender@planetexpress.com`
2. Setup SAML with an user `uid=bender@planetexpress.com` and attribute to map user to existing LDAP `users = mail`
3. Login with SAML, we will end up with the LDAP user

### DIFF
--- a/lib/UserResolver.php
+++ b/lib/UserResolver.php
@@ -9,19 +9,52 @@ declare(strict_types=1);
 namespace OCA\User_SAML;
 
 use OCA\User_SAML\Exceptions\NoUserFoundException;
+use OCP\IConfig;
 use OCP\IUser;
 use OCP\IUserManager;
+use OCP\LDAP\Exceptions\MultipleUsersReturnedException;
+use OCP\LDAP\ILDAPProviderFactory;
+use OCP\Server;
 
 class UserResolver {
 	public function __construct(
-		private IUserManager $userManager,
+		private readonly IUserManager $userManager,
+		private readonly ILDAPProviderFactory $ldapProviderFactory,
+		private readonly IConfig $config,
 	) {
 	}
 
 	/**
 	 * @throws NoUserFoundException
 	 */
-	public function findExistingUserId(string $rawUidCandidate, bool $force = false, bool $isActiveDirectory = false): string {
+	public function findExistingUserId(string $rawUidCandidate, ?array $idpSettings = null, bool $force = false, bool $isActiveDirectory = false): string {
+		// If configured, find the user based on a different LDAP attribute.
+		if ($idpSettings !== null
+			&& version_compare($this->config->getSystemValueString('version', '0.0.0'), '34.0.0', '>=')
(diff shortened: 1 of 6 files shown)


lib/UserResolver.php
```diff
@@ -9,19 +9,49 @@
...
+					$user = $this->ldapProviderFactory->getLDAPProvider()->findOneUserByAttributeValue($userIdLdapMapping, $rawUidCandidate);
+				}
+			} catch (MultipleUsersReturnedException $e) {
+				return '';
+			}
+			if ($user === null) {
+				return '';
+			}
+
+			return $user->getUID();
```
CarlSchwan:
```suggestion
			} catch (MultipleUsersReturnedException $e) {
				$user = null;
			}
			if ($user !== null) {
				return $user->getUID();
			}
			// fallback to normal search
```

lib/UserResolver.php
```diff
@@ -9,19 +9,49 @@
...
 	/**
 	 * @throws NoUserFoundException
 	 */
-	public function findExistingUserId(string $rawUidCandidate, bool $force = false, bool $isActiveDirectory = false): string {
+	public function findExistingUserId(string $rawUidCandidate, array $idpSettings = null, bool $force = false, bool $isActiveDirectory = false): string {
+		// If configured, find the user based on a different LDAP attribute.
+		if ($idpSettings !== null
+			&& version_compare($this->config->getSystemValueString('version', '0.0.0'), '34.0.0', '>=')
+			&& $this->ldapProviderFactory->isAvailable()
+			&& isset($idpSettings['saml-attribute-mapping-user_id_ldap_mapping'])) {
```
come-nc: Is the isset check enough? Can it not be an empty string if it was left empty?

lib/UserResolver.php
come-nc: I’m not sure you should fallback to normal search, especially if multiple users were found.
What is the rationale for falling back? It encourages bad configurations IMO.

CarlSchwan: Agree about the MultipleUsersReturnedException,

But when no user is found we need to continue the regular procedure, according to https://github.com/nextcloud/server/issues/55284 

> Other login backends that may rely on LDAP, like the SAML/SSO or OIDC backend, will get an additional configuration option to specify an attribute to which the received user id is being compared. When an admin configured it, and an LDAPProvider is available, upon login the backend will try to resolve the user with the afore mentioned method. If no user was found the regular procedure continues, if more than one user were found (exception case) it will be logged accordingly, and the Not Provided error will be shown (background: prevent data leakage).

The main usecase of this feature is to be able to create a custom ldap containing the old id of an user (e.g. previousEmail) and then SAML will match against this old id fist.

come-nc: Yeah ok the old-id usecase makes sense.

come-nc (approved):

PR #1132, opened 2026-06-01
8.1.2

CarlSchwan:


blizzz (approved):
what's the root cause? DI hell?

AndyScherzinger (approved):

CarlSchwan:
> what's the root cause? DI hell?

the event triggers a listener in Guest which expect the session to be set. But it seems that while we setup the session just before triggering the listener it is only fully set after the redirect. I need to investigate this a bit more, but reverting the commit fixes the issue for now

AndyScherzinger (approved):

PR #1137, opened 2026-06-08
fix: Update display name in User object

artonge:
This is breaking other consumers of the user object as they will use the old display name. This was an issue in Talk where they cache the display names.

- Need https://github.com/nextcloud/server/pull/61593

### DIFF
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -541,7 +541,7 @@ class UserBackend extends ABackend implements IApacheBackend, IUserBackend, IGet
 			$currentDisplayname = $this->getDisplayName($uid);
 			if ($newDisplayname !== null
 				&& $currentDisplayname !== $newDisplayname) {
-				$this->setDisplayName($uid, $newDisplayname);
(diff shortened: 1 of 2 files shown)


lib/UserBackend.php
```diff
@@ -542,6 +542,7 @@ public function updateAttributes(string $uid): void {
 			if ($newDisplayname !== null
 				&& $currentDisplayname !== $newDisplayname) {
 				$this->setDisplayName($uid, $newDisplayname);
+				$user->setDisplayName($newDisplayname);
```
artonge: This could be enough as I suspect that `$user->setDisplayName` will eventually call `$this->setDisplayName`.

```suggestion
				$user->setDisplayName($newDisplayname);
```

CarlSchwan: :+1:

artonge: I double-checked, and server needed some fix in order for `$this->setDisplayName` to be called through `$user->setDisplayName`

CarlSchwan (approved):

PR #1138, opened 2026-06-10
fix: graceful handling when user_saml is enabled but no IdP is configured

printminion-co:
## Summary

- Fixes a **500 TypeError** that occurs when `user_saml` is active (`type=saml`) but no IdP has been configured
- Redirects to the existing `genericError` page with a translatable user-readable message instead of crashing
- Adds `occ saml:config:validate` for one-shot admin/support diagnostics (exit 0/1/2)

## Reproduction

1. Install/enable the `user_saml` app
2. `occ config:app:set user_saml type --value saml`
3. Do **not** add any IdP via the admin panel
4. Open the Nextcloud login page → **500 TypeError**

## Root Cause

The crash is a two-step interaction within the same request:

1. `Application.php` beforeware calls `$samlSettings->getListOfIdps()` → `ensureConfigurationsLoaded(-1)` loads from DB → DB is empty → sets `configurationsLoadedState = LOADED_ALL`
2. `SAMLController::login(1)` calls `getOneLoginSettingsArray(1)` → `ensureConfigurationsLoaded(1)` returns early (already `LOADED_ALL`) → `$this->configurations[1]` is never set → remains `null`
3. Line 156 calls `array_key_exists('sp-entityId', $this->configurations[$idp])` — `array_key_exists()` on `null` throws a `TypeError`

Every other field access in `getOneLoginSettingsArray()` uses the safe `?? ''` null-coalesce pattern; line 156 was the one outlier using `array_key_exists()`.

### `lib/SAMLSettings.php`
- Replace the `array_key_exists()` call with `?? ''` (consistent with all other lines in the method)
- Add an early guard after `ensureConfigurationsLoaded()` that throws `InvalidArgumentException` when the requested IdP is absent from `$configurations`, converting a cryptic null-pointer into a clear, actionable message

### `lib/Controller/SAMLController.php`
- Catch `InvalidArgumentException` from `getOneLoginSettingsArray()` in `login()` and redirect to the existing `genericError` page with a translatable user-readable message, instead of letting the exception propagate as a 500

### `occ saml:config:validate` (new command)

Gives support teams and administrators a single command to diagnose the SAML configuration state without reading raw database rows:

```
$ occ saml:config:validate
Type: saml
  IdP #1 (My IdP): MISSING idp-entityId, idp-singleSignOnService.url
```

Exit codes:

| Code | Meaning |
|------|---------|
| `0` | All configured IdPs pass required-field validation |
| `1` | `type` not set — app is not active |
| `2` | Active but one or more required fields are missing |

Validated fields per IdP: `idp-entityId`, `idp-singleSignOnService.url`, `general-uid_mapping`.

New files: `lib/Command/ConfigValidate.php`, `tests/unit/Command/ConfigValidateTest.php`, + one line in `appinfo/info.xml`.

## Tests

- **`tests/unit/SAMLSettingsTest.php`** (new): reproduces the exact `LOADED_ALL` bug scenario; also covers the simple not-in-DB case, happy path, and `sp-entityId` fallback to metadata URL
- **`tests/unit/Controller/SAMLControllerTest.php`**: adds `testLoginWithUnconfiguredIdpRedirectsToGenericError()` — verifies the controller redirects to `genericError` when `getOneLoginSettingsArray()` throws `InvalidArgumentException`
- **`tests/unit/Command/ConfigValidateTest.php`** (new): covers all four exit-code paths (type not set, no IdPs, OK, missing fields)

## After This Fix

- `/login` — shows the normal Nextcloud login page (no crash; admins can still log in to fix the configuration)
- `/apps/user_saml/saml/login?idp=1` — redirects to the existing `genericError` page with a readable message: _"SAML authentication is not configured. Please ask your administrator to complete the SAML setup in the admin panel."_
- `data/nextcloud.log` — no `array_key_exists(): Argument #2 must be of type array, null given`

### DIFF
--- a/lib/Controller/SAMLController.php
+++ b/lib/Controller/SAMLController.php
@@ -161,7 +161,16 @@ class SAMLController extends Controller {
 		$type = $this->appConfig->getAppValueString('type');
 		switch ($type) {
 			case 'saml':
-				$settings = $this->samlSettings->getOneLoginSettingsArray($idp);
+				try {
+					$settings = $this->samlSettings->getOneLoginSettingsArray($idp);
+				} catch (\InvalidArgumentException $e) {
+					return new Http\RedirectResponse(
+						$this->urlGenerator->linkToRouteAbsolute(
+							'user_saml.SAML.genericError',
+							['reason' => 'notConfigured']
+						)
+					);
+				}
 				$auth = new Auth($settings);
 				$passthroughParamsString = trim($settings['idp']['passthroughParameters'] ?? '') ;
 				$passthroughParams = array_map(trim(...), explode(',', $passthroughParamsString));
@@ -555,6 +564,7 @@ class SAMLController extends Controller {
 		$allowedMessages = [
 			'userDisabled' => $this->l->t('This user account is disabled, please contact your administrator.'),
 			'authFailed' => $this->l->t('Authentication failed.'),
+			'notConfigured' => $this->l->t('SAML authentication is not configured. Please ask your administrator to complete the SAML setup in the admin panel.'),
 		];
 
 		$message = $allowedMessages[$reason] ?? $this->l->t('Unknown error, please check the log file for more details.');
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -57,6 +57,7 @@ While theoretically any other authentication provider implementing either one of
 		<command>OCA\User_SAML\Command\ConfigDelete</command>
 		<command>OCA\User_SAML\Command\ConfigGet</command>
 		<command>OCA\User_SAML\Command\ConfigSet</command>
+		<command>OCA\User_SAML\Command\ConfigValidate</command>
 		<command>OCA\User_SAML\Command\GetMetadata</command>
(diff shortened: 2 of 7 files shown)


lib/Controller/SAMLController.php
```diff
@@ -161,7 +161,16 @@ public function login(int $idp = 1): Http\RedirectResponse|Http\TemplateResponse
...
 		switch ($type) {
 			case 'saml':
-				$settings = $this->samlSettings->getOneLoginSettingsArray($idp);
+				try {
+					$settings = $this->samlSettings->getOneLoginSettingsArray($idp);
+				} catch (\InvalidArgumentException $e) {
+					return new Http\RedirectResponse(
+						$this->urlGenerator->linkToRouteAbsolute(
+							'user_saml.SAML.genericError',
+							['message' => $this->l->t('SAML authentication is not configured. Please ask your administrator to complete the SAML setup in the admin panel.')]
```
CarlSchwan: This does not work anymore. You need to add the  message in $allowedMessages

CarlSchwan (approved):

PR #1216, opened 2026-08-21
fix: searching in groups

CarlSchwan:
And add unit tests

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI


julien-nc (approved):
Did not test myself but got feedback that it fixes the issue in a real setup.

PR #1219, opened 2026-08-27
feat: Implement ISearchKnownUsersBackend

CarlSchwan:
## 🤖 AI (if applicable)

### DIFF
--- a/lib/UserBackend.php
+++ b/lib/UserBackend.php
@@ -31,6 +31,7 @@ use OCP\User\Backend\IGetDisplayNameBackend;
 use OCP\User\Backend\IGetHomeBackend;
 use OCP\User\Backend\ILimitAwareCountUsersBackend;
 use OCP\User\Backend\IProvideEnabledStateBackend;
+use OCP\User\Backend\ISearchKnownUsersBackend;
 use OCP\User\Backend\ISetDisplayNameBackend;
 use OCP\User\Events\UserChangedEvent;
 use OCP\User\Events\UserFirstTimeLoggedInEvent;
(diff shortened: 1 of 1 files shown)


lib/UserBackend.php
```diff
@@ -31,14 +31,18 @@
...
 use OCP\User\Backend\ISetDisplayNameBackend;
 use OCP\User\Events\UserChangedEvent;
 use OCP\User\Events\UserFirstTimeLoggedInEvent;
 use OCP\UserInterface;
 use Override;
 use Psr\Log\LoggerInterface;
 
-class UserBackend extends ABackend implements IApacheBackend, IUserBackend, IGetDisplayNameBackend, ILimitAwareCountUsersBackend, IGetHomeBackend, ICustomLogout, ISetDisplayNameBackend, IProvideEnabledStateBackend {
+class UserBackend extends ABackend implements IApacheBackend, IUserBackend, IGetDisplayNameBackend, ILimitAwareCountUsersBackend, IGetHomeBackend, ICustomLogout, ISetDisplayNameBackend, IProvideEnabledStateBackend, ISearchKnownUsersBackend {
+	private const table = 'user_saml_users';
```
cristianscheid, line 43: since this const is being introduced, maybe worth also using it on other methods instead of plain string 'user_saml_users'?

cristianscheid (approved):
only a nitpick, other than that looks good to me

lib/UserBackend.php
CarlSchwan: I wanted to do it as a follow up, but I can just do it as seperate commit to this PR

come-nc (approved):

PR #1228, opened 2026-09-08
feat(admin): Display list of available attributes

CarlSchwan:
Make it easier for admin to discover that is available

## 🤖 AI (if applicable)


lib/SAMLSettings.php
```diff
@@ -222,6 +231,31 @@ public function set(int $id, array $settings): void {
...
+	}
+
+	/**
+	 * @param list<string> $attributeKeys
+	 */
+	public function recordAvailableAttributes(int $id, array $attributeKeys): void {
+		$cacheKey = (string)$id;
+		$known = $this->knownAttributesCache->get($cacheKey) ?? [];
+		$updated = array_values(array_unique(array_merge($known, $attributeKeys)));
+		if (count($updated) > self::MAX_KNOWN_ATTRIBUTES) {
```
cristianscheid, line 246: maybe we could log something on this case?

cristianscheid (approved):
only one minor suggestion, other than that LGTM

lib/SAMLSettings.php
CarlSchwan: But that would mean logging for every logging. Maybe just a info or debug message, so per default it doesn't log anything

cristianscheid: Oh ok, I didn't realize this would log on every login. The idea was just so admins are aware when attributes are being capped, so I agree with you, maybe just a info or debug message, if that makes sense.

Issue #1194, opened 2026-07-22
Investigate query count


CarlSchwan:
`Controller OCA\\User_SAML\\Controller\\SAMLController::assertionConsumerService executed 2460 queries.` found in my log

CarlSchwan:
Done :)

PR #1231, opened 2026-09-10
perf: Don't fetch all users from a group on login

CarlSchwan:
Instead of fetching all users from a group to then check if they are in the saml backend. Do directly a join on the saml user table.

This should fix the multiple thousands of DB requests created on any SAML logins.

## 🤖 AI (if applicable)


come-nc (approved):

CarlSchwan:
Fix #1194 btw

PR #1235, opened 2026-09-15
fix(logout): Ensure we handle kerberos correctly

CarlSchwan:
Where logout button doesn't make sense

Works best with https://github.com/nextcloud/server/pull/64343 

## 🤖 AI (if applicable)

- [ ] The content of this PR was partly or fully generated using AI (N/A)


lib/UserBackend.php
```diff
@@ -271,13 +271,20 @@ public function getLogoutUrl(): string {
...
 
-		$logoutUrl = $this->urlGenerator->linkToRouteAbsolute($slo === '' ? 'core.login.logout' : 'user_saml.SAML.singleLogoutService', [
-			'requesttoken' => $tokenManager->getToken()->getEncryptedValue(),
-		]);
-		/** @var non-empty-string $logoutUrl */
-		return $logoutUrl;
+		// Without the normal login form, /login redirects straight back into SAML,
+		// so a local logout would be pointless.
+		if ($slo === '' && !$this->appConfig->getAppValueBool('general-nextcloud_login_form')) {
+			return '';
```
blizzz, line 278: Maybe I am not seeing something, but this will still cause the same `core.login.logout` route, no? The server change does not, eh, change anything in this regard.

I also do not see any specific relation to Kerberos in the change.

grnd-alt: @blizzz server pr is updated now, to hide the login button if an empty string is returned from the user backend

(kerberos relation is that kerberos doesn't provide a logout url)

blizzz: > (kerberos relation is that kerberos doesn't provide a logout url)

The logout URL is a configuration bit, a built-in SAML one does not have to provide one either.

CarlSchwan: With a build-in SAML, we should be able to logout got the normal nextcloud login form and either login with another SAML provider or another login method

blizzz: > With a build-in SAML, we should be able to logout got the normal nextcloud login form and either login with another SAML provider or another login method

With the own logout form, it redirects to the login page, which would redirect to the SSO page. When the session there is still active, it would redirect back into Nextcloud and the user would be logged in again. And if not they would stay with the SSO provider. All provided SAML is the only login option (so many conditions). Thus in some cases questionable user experience, but it would work. Tbh I do not recall the current behaviour when SLO is not configured :speak_no_evil: 

Having this not bound to kerberos is not an issue per se, the commit message is then misleading though and should be corrected. Think about your and our future selves!

